PR #2, opened 2016-12-07
versions update

KadekM:


LukaJCB:
Thanks, was just about to do this!

Issue #16, opened 2017-03-03
Refactor tags and attributes


PerWiklander:
As a user of the lib I would like to be able to better namespace tags and attributes and only import what is needed.

I would also like to have access to documentation showing me what the various tags and attributes are used for.

PR coming soon.

PR #18, opened 2017-03-06
#16 The missing parts

PerWiklander:
There were a few commits that didn't make it into the previous PR.


PerWiklander:
Is it possible to kill the merge commit you made before? Resolving the conflict is... yucky.

PR #27, opened 2017-04-09
Improve Store code readability

guizmaii:
Hi,

IMHO, `Store[T, U]` is not very readable and so not very easy to understand for newbies, like me, of ELM architecture so I improved it a little. 
And I added return types because it could also help newbies to understand what your code do and why it's useful.


LukaJCB:
This looks great! Thanks for your contribution!

Issue #32, opened 2017-05-24
Tags and Attributes namespace


mariusmuja:
Importing all the tags and attributes into the global namespace can be the cause of confusing error messages and IDE refactoring issues. 

I suggest using the same approach as in the scalajs-react vdom, by putting the tags into a ```<``` namespace and the attributes into a ```^``` namespace, while offering the option of importing everything into the global namespace if the user wants to.

gaydenko:
To my taste both `<` and `^` are very ugly. But some namespaces separation can be useful, agree. May be to use something else (say `tags`, `atts`), and react-addicted users will be able to rename objects while importing.

mariusmuja:
Not a big fan of ```<``` and ```^``` either, what I would usually do is import the all the tags and attributes, but not globally, just in the function creating the vdom.

The reason I suggested ```<``` and ```^``` was to keep it similar to scalajs-react. ```tags``` and ```attrs``` are more reasonable names, although longer to type, so I imagine most users would rename on import. Would it make sense to have something like ```T``` and ```A``` available as well (they would be unlikely to cause collisions)?

gaydenko:
I don't understand why just "similar to scalajs-react" can be a reason to follow :) 
Yes, `A` and `T` as default objects' names is also acceptable, of course. Anything not breaking eyes feelings.

mariusmuja:
Updated PR to use ```T``` and ```A```.

LukaJCB:
This is a good idea, Although we don't need to break backwards compatability for that. We could just add an Object that contains all the Tags and one for the Attributes and using import aliases we could name them whatever we want. I.e. `T` or `<` or `^` or anything else.

```
import outwatch.dom.{Attributes => ^, Tags => <}

val node = <.div(^.className := "some-class")
```

Or 

```
import outwatch.dom.{Attributes => A, Tags => T}

val node = T.div(A.className := "some-class")
```

LukaJCB:
I just pushed a commit (a355bde2026f9253982ad8c2815146ee5fe2af1e) that fixes this issue and will be releasing it in version 0.9.5 :)

mariusmuja:
Looks good. How about adding an object ```all``` in the dom package that contains all the tags and attributes for having a single import in a local namespace (such as a function creating a vnode)?

mariusmuja:
> Looks good. How about adding an object all in the dom package that contains all the tags and attributes for having a single import in a local namespace (such as a function creating a vnode)?

[...]

PR #33, opened 2017-05-24
Namespaces for tags and attributes

mariusmuja:
An attempt at fixing #32.

Tags and attributes are imported namespaced by default with ```import outwatch.dom._``` and globally with ```import outwatch.dom.all._```.

Note, this will break most client code, it will require adding ```import outwatch.dom.all._``` to keep the same code working.


LukaJCB:
Hey, thank you so much for your contribution! However, I think this should be done without breaking  compatibility :)

Issue #38, opened 2017-06-23
Snabbdom patching issue


mariusmuja:
I came across an issue while trying to build a router for outwatch. In some cases when a "page" was replaced with itself (but different initial state), the observable updates on the new page would not work.

I tracked the issue to the fact that Snabbdom patching does an update (and not a destroy/insert) which results in outwatch not subscribing to the new observables, since the subscription in done in the insert hook.

Adding the key property didn't fix the issue, it revealed a second bug caused by outwatch not passing the key to the ```DataObject``` constructed by ```DataObject.updateAttributes```. 

I fixed these issues on my branch, PR to follow.

The snippet below shows the issue, going from page 1 -> 3 or 3 ->1 breaks the counter updates, but all other transitions (1->2, 3->2,...) work as expected:

```scala
  def page(num: Int): VNode = {
    val clicks = createHandler[Int]()
    val sum = clicks.startWith(num)
      .scan(0)(_ + _)

    div(
      num match {
        case 1 =>
          div(child <-- sum)
        case 2 =>
          textarea(child <-- sum)
        case 3 =>
          div(child <-- sum)
      },
      div(
        button(click(1) --> clicks, "+1")
      )
    )
  }

  def root() : VNode = {

    val pageHandler = createHandler[Int](1)

    div(
      button(click(1) --> pageHandler, "Page 1"),
      button(click(2) --> pageHandler, "Page 2"),
      button(click(3) --> pageHandler, "Page 3"),
      div(child <-- pageHandler.map(page))
    )
  }

  def main(): Unit = {
    OutWatch.render("#app", root())
  }
```

Issue #7, opened 2016-12-16
Occasional weirdness in DOM updates


KadekM:
To reproduce:

https://github.com/KadekM/outwatch-playground/blob/master/src/main/scala/Outwatchplayground.scala

It seems to reuse the same observable (not attach new one), which seems to cause mentioned problems...

raquo:
I've just started wrapping my head around Outwatch's internals, but I think a possible reason for this is that Outwatch creates the subscription in snabbdom's `insert` hook, which fires only when a new DOM element is created (not when a new _VNode_ is created). So if snabbdom chooses to reuse the same DOM node, which is probably the case here, the insert hook will not fire. Same for destroy hook. And as you said, we're left with the old subscription.

pme123:
I have the same or similar problem:
`children <-- Observable.from(Future(uiEntity.parameterEdit()))`
uiEntity.parameterEdit() creates a Seq of VNodes. The uiEntity changes, but in the UI the parent element shows still the first VNodes, that were shown the first time (when loading).
Btw: this is the only way I saw to add a Seq of VNodes. Why is something like
`children := uiEntity.parameterEdit()` not possible?

LukaJCB:
Hey @pme123, yes, that is indeed confusing. However it is possible. Instead of doing `children := ...` you can expand a `Seq` into varargs by annotating them with `: _*`. So something like `div(uiEntity.parameterEdit(): _*)`

pme123:
thanx, that was exactly what I was looking for. (BTW: great project!)

PR #39, opened 2017-06-23
Fix Snabbdom patching issue

mariusmuja:
This PR fixes issue #38.

The patch assigns a key to a VNode's data object that contains changeables and doesn't already have a key in order to force Snabbdom to destroy/re-create it when patched with new changeables. It also fixes the ```DataObject.updateAttributes``` method by passing the key to the newly constructed ```DataObject```.

I've also been doing some refactoring, most notably turned DataObject and Hooks in the Snabbdom facade into ```@ScalaJSDefined``` traits for extra type safety (this would have prevented the ```DataObject.updateAttributes``` bug for example).

### DIFF
--- a/src/main/scala/outwatch/dom/VDomModifier.scala
+++ b/src/main/scala/outwatch/dom/VDomModifier.scala
@@ -45,8 +45,7 @@ object VDomModifier {
 
   final case class VTree(nodeType: String,
                    children: Seq[VNode],
-                   attributeObject: DataObject,
-                   changables: Observable[(Seq[Attribute], Seq[VNode])]
+                   attributeObject: DataObject
                   ) extends VNode {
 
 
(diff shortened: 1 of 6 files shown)


KadekM:
I wonder if this could potentially help with #7 ?

LukaJCB:
Wow! That's really awesome! Thank you so so much! 

The only small thing I'd like to add is a test case that tests this behaviour, we could probably use the minimal test case you gave in #38. I can add it myself if you're too busy.

Again, thank you so much!

mariusmuja:
I added a test case for this behaviour.

mariusmuja:
@KadekM Yes, this fixes #7, it's the same issue.

LukaJCB:
Thank you again, really appreciate this! I'll make a release later today!

mariusmuja:
@LukaJCB You're welcome! Thank you for outwatch, it's an awesome project.

LukaJCB:
Version 0.10.1 is live! :)

PR #41, opened 2017-06-28
Some refactorings in Sink

mariusmuja:
A few refactorings in ```Sink```, made a code a bit more DRY, added ```onError``` and ```onComplete``` arguments to ```Sink.create```.

### DIFF
--- a/src/main/scala/outwatch/Sink.scala
+++ b/src/main/scala/outwatch/Sink.scala
@@ -1,8 +1,11 @@
 package outwatch
 
+import outwatch.dom.Handler
 import rxscalajs.facade.SubjectFacade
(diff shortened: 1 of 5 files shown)


LukaJCB:
Hey @mariusmuja! Great work! Really appreciate it! However, I'm not sure why exactly we'd need `onError` and `onComplete` arguments. The streams used for Sinks are usually infinite and don't include errors, no?

mariusmuja:
@LukaJCB Yes, that's true. However Sinks can also be used for network communication (websockets for example), in which case network errors can occur. Also, the ```onError``` and ```onComplete``` arguments have default values, so the ```Socket.create``` usage remains the same in most cases.

PR #47, opened 2017-09-23
Introduce Style Attribute

fdietze:
Hi, we introduced a Style attribute which uses the Style Module of Snabbdom. Is this going in the right direction? Should we make it usable like ScalaTags?

### DIFF
--- a/src/main/scala/snabbdom/VDomProxy.scala
+++ b/src/main/scala/snabbdom/VDomProxy.scala
@@ -14,15 +14,18 @@ object VDomProxy {
 
   import js.JSConverters._
 
-  def attrsToSnabbDom(attributes: Seq[Attribute]): (js.Dictionary[String], js.Dictionary[String]) = {
-    val (attrs, props) = attributes.partition {
-      case (a: Attr) => true
-      case (p: Prop) => false
-    }
-    val attrDict = js.Dictionary(attrs.map(attr => attr.title -> attr.value): _*)
-    val propDict = js.Dictionary(props.map(prop => prop.title -> prop.value): _*)
+  def attrsToSnabbDom(attributes: Seq[Attribute]): (js.Dictionary[String], js.Dictionary[String], js.Dictionary[String]) = {
+    val attrsDict = js.Dictionary[String]()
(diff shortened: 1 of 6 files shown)


fdietze:
How did you port the MDN data for the tags and attributes? Is it possible to do the same for Styles?

src/main/scala/snabbdom/VDomProxy.scala
```diff
@@ -14,15 +14,18 @@ object VDomProxy {
...
-  def attrsToSnabbDom(attributes: Seq[Attribute]): (js.Dictionary[String], js.Dictionary[String]) = {
-    val (attrs, props) = attributes.partition {
-      case (a: Attr) => true
-      case (p: Prop) => false
+  def attrsToSnabbDom(attributes: Seq[Attribute]): (js.Dictionary[String], js.Dictionary[String], js.Dictionary[String]) = {
+    val attrsDict = js.Dictionary[String]()
+    val propsDict = js.Dictionary[String]()
+    val styleDict = js.Dictionary[String]()
+
+    attributes.foreach {
```
LukaJCB: Can you change this to a fold? :)

fdietze: I really thought about doing this with a fold. But I didn't come up with a way that has the same performance characteristics as my current solution. With building up a triple of lists inside a fold and constructing dictionaries afterwards, we have to iterate twice over every element. We cannot build an immutable triple of dictionaries. And in the end we have to write it into a mutable dictionary anyways: https://github.com/scala-js/scala-js/blob/master/library/src/main/scala/scala/scalajs/js/Dictionary.scala#L59

So I think this is a trade-off between being pseudo-pure and performance. I chose performance. Do you still prefer a fold? ;)

PerWiklander:
> fdietze commented 3 days ago:
> How did you port the MDN data for the tags and attributes? Is it possible to do the same for Styles?

I copied most of it from ScalaTags and looked the rest up in MDN. I believe @lihaoyi had a script to do the bulk of the extraction in the beginning.

PR #48, opened 2017-09-25
support optional elements and attributes

### LINKED ISSUE #22: Optional elements and attributes
This would make it easier to include an element or not depending on if an `Option` (or `js.UndefOr`) is defined or not (taken from scalajs-react).
This would work with implicit conversion,  like `StringNode` today.

```scala
case class User(
  profileUrl: String,
  favColor: Option[String]
)

val loggedInUser: Option[User] = ???

div(
  h3("Welcome"),
  loggedInUser.map(user =>
    a(
      href := user.profileUrl,
      style := user.favColor.map(c => s"background-color: $c")
      "My Profile"
    )
  )
)
```

cornerman:
This is my attempt to fix #22. Does it make sense to you?

### DIFF
--- a/src/main/scala/outwatch/dom/VDomModifier.scala
+++ b/src/main/scala/outwatch/dom/VDomModifier.scala
@@ -45,13 +45,15 @@ final case class AttributeStreamReceiver(attribute: String, attributeStream: Obs
 final case class ChildStreamReceiver(childStream: Observable[VNode]) extends Receiver
 final case class ChildrenStreamReceiver(childrenStream: Observable[Seq[VNode]]) extends Receiver
 
-
+final case object EmptyVDomModifier extends VDomModifier
 
 
 object VDomModifier {
   final implicit class StringNode(string: String) extends VNode {
     val asProxy = VNodeProxy.fromString(string)
   }
+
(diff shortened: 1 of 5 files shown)


LukaJCB:
LGTM, thank you very much :)

PerWiklander:
Nice. Thanks!
Did we really need the new operator `:=?` though?

cornerman:
In the corresponding issue I have looked at the [link from scalajs-react](https://github.com/japgolly/scalajs-react/blob/master/doc/VDOM.md#conditional-vdom) and implemented it with the same operator, which makes the semantics explicit. Overloading the `:=` operator would also be fine with me; but trying it now, I am also not really able to convince the compiler to use the option overload :)

PerWiklander:
Being explicit might be a good idea actually. But... wouldn't having `OptionIsEmptyModifier` in scope make it work with `:=` ? So that I could choose to import that if I want to use the normal operator and be implicit?

cornerman:
Well, the implicit is already in scope, you can already write it like so:
```
input(
  Option("bar").map(data.foo := _),
  Option.empty.map(data.bar := _)
)
```

However, I also tried importing the implicit but did not help. Ideas welcome!

PerWiklander:
Sorry, I slipped while thinking :-)

PR #58, opened 2017-10-06
Refactor to referential transparent api

LukaJCB:
First version where all tests pass!
Needs more work on `Store` and the like. For now, though, I'm quite happy how it turned out.

Will also have to update the documentation when this is merged (I'd like to use `tut` in the future for docs) :)

### DIFF
--- a/src/main/scala/outwatch/dom/helpers/DomUtils.scala
+++ b/src/main/scala/outwatch/dom/helpers/DomUtils.scala
@@ -25,10 +25,13 @@ object DomUtils {
         childStreamReceivers.map(_.childStream)
       )
 
-      val childrenReceivers = childrenStreamReceivers.headOption.map(_.childrenStream)
+      val childrenReceivers = childrenStreamReceivers.lastOption.map(_.childrenStream)
 
       val attributeReceivers: Observable[Seq[Attribute]] = Observable.combineLatest(
-        attributeStreamReceivers.map(_.attributeStream)
+        attributeStreamReceivers
+          .groupBy(_.attribute)
+          .values
+          .map(_.last.attributeStream)(breakOut)
       )
 
       val allChildReceivers = childrenReceivers.getOrElse(childReceivers)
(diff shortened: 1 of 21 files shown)


src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -123,65 +124,64 @@ object DomUtils {
...
+    ()
   }
 
 
   private[outwatch] final case class SeparatedModifiers(
-    emitters: Seq[Emitter] = Seq.empty[Emitter],
-    receivers: Seq[Receiver] = Seq.empty[Receiver],
-    properties: Seq[Property] = Seq.empty[Property],
-    vNodes: Seq[VNode] = Seq.empty[VNode]
+    emitters: List[Emitter] = List.empty[Emitter],
```
fdietze: Maybe just write `Nil` here? It is `Nil` anyways:https://github.com/scala/scala/blob/v2.12.3/src/library/scala/collection/immutable/List.scala#L463

And maybe move this change into a separate commit.

src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -25,10 +25,13 @@ object DomUtils {
         childStreamReceivers.map(_.childStream)
       )
 
-      val childrenReceivers = childrenStreamReceivers.headOption.map(_.childrenStream)
+      val childrenReceivers = childrenStreamReceivers.lastOption.map(_.childrenStream)
 
       val attributeReceivers: Observable[Seq[Attribute]] = Observable.combineLatest(
-        attributeStreamReceivers.map(_.attributeStream)
+        attributeStreamReceivers
```
fdietze: Add a comment to make clear why we are doing this:
```scala
// only use last encountered observable per attribute
```

src/test/scala/outwatch/LifecycleHookSpec.scala
```diff
@@ -34,6 +36,61 @@ class LifecycleHookSpec extends UnitSpec with BeforeAndAfterEach {
...
+    OutWatch.render("#app", node).unsafeRunSync()
+    switch1 shouldBe false
+    switch2 shouldBe false
+
+    message.next("wursi")
+    switch1 shouldBe true
+    switch2 shouldBe true
+  }
+
   "Destruction hooks" should "be called correctly" in {
```
fdietze: move to other `Destruction hooks` and use `it should ...`
(same for Insertion and Update hooks)

PR #60, opened 2017-10-07
Delayed asProxy, VTree.apply to copy with more modifiers

fdietze:
also fixes overwriting of StreamReceivers:
e.g.
```scala
div(data <-- a, data <-- b) // only uses stream b instead of both
div(children <-- a, children <-- b) // only uses last stream b instead of first
```

It makes sense to merge this PR before the referential transparency one, because constructing `VNode` is not a side-effect anymore.

### DIFF
--- a/src/main/scala/outwatch/dom/VDomModifier.scala
+++ b/src/main/scala/outwatch/dom/VDomModifier.scala
@@ -18,7 +18,7 @@ sealed trait Property extends VDomModifier
 
 sealed trait Receiver extends VDomModifier
 
-sealed trait VNode extends VDomModifier {
+sealed trait VNode extends Any with VDomModifier {
   def asProxy: VNodeProxy
 }
 
@@ -51,7 +51,7 @@ final case object EmptyVDomModifier extends VDomModifier
 
 
 object VDomModifier {
-  final implicit class StringNode(string: String) extends VNode {
+  final implicit class StringNode(val string: String) extends AnyVal with VNode {
     def asProxy = VNodeProxy.fromString(string)
   }
 
--- a/src/main/scala/outwatch/dom/VDomModifier.scala
+++ b/src/main/scala/outwatch/dom/VDomModifier.scala
@@ -3,7 +3,7 @@ package outwatch.dom
 import org.scalajs.dom._
 import scala.scalajs.js.|
 import rxscalajs.{Observable, Observer}
-import snabbdom.{DataObject, VNodeProxy, h}
+import snabbdom.{VNodeProxy, h}
 
 import scala.scalajs.js
 import collection.breakOut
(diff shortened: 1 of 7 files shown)


src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -51,25 +51,21 @@ final case object EmptyVDomModifier extends VDomModifier
...
+    def asProxy = VNodeProxy.fromString(string)
   }
 
   implicit def OptionIsEmptyModifier(opt: Option[VDomModifier]): VDomModifier = opt getOrElse EmptyVDomModifier
 
-  final case class VTree(nodeType: String,
-                   children: Seq[VNode],
-                   attributeObject: DataObject
-                  ) extends VNode {
+  final case class VTree(nodeType: String, modifiers: Vector[VDomModifier]) extends VNode {
```
LukaJCB: Why make this a vector? I think for small collections like this it makes more sense to have a `List`, no?

src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -51,25 +51,21 @@ final case object EmptyVDomModifier extends VDomModifier
...
+      val (children, attributeObject) = DomUtils.extractChildrenAndDataObject(modifiers)
+      val childProxies: js.Array[VNodeProxy] = children.map(_.asProxy)(breakOut)
+      h(nodeType, attributeObject, childProxies)
+    }
 
-    lazy val childProxies: js.Array[VNodeProxy] = children.map(_.asProxy)(breakOut)
-
-    val asProxy = h(nodeType, attributeObject, childProxies)
-
+    def apply(args: VDomModifier*):VNode = VTree(nodeType, modifiers ++ args)
```
LukaJCB: Can we move this to `VNode` so we don't have to change all the type signatures? We could just pattern match, and it could be a NoOp on `StringNode`.

src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -51,7 +51,7 @@ final case object EmptyVDomModifier extends VDomModifier
 
 
 object VDomModifier {
-  final implicit class StringNode(string: String) extends VNode {
+  final implicit class StringNode(val string: String) extends AnyVal with VNode {
```
LukaJCB: 👍

src/main/scala/outwatch/dom/VDomModifier.scala
fdietze: `apply()` on `StringNode` does not make sense, so I think it is better to encode that in the type. Having that in `VNode` and a NoOp on `StringNode` could lead to confusion.

src/main/scala/outwatch/dom/VDomModifier.scala
fdietze: `Vector` has better complexity for concatenation and `lastOption`. But for these short lists we need a benchmark on the javascript runtime. Maybe arrays are the fastest. Arrays would also have the advantage that we could directly store the passed arguments without conversion.

PR #68, opened 2017-10-30
Never hide IO Monad in API

fdietze:
This PR does not hide the IO Monad anymore. This means that the following code:
```scala
val pureComponent:VNodeIO = div(...)
val component:VNodeIO = for {
  handler <- createHandler
  div <- div( ... handler ... )
} yield div
val bigComponent = div(component, purecomponent)
```
now needs to be written like:

```scala
val pureComponent:VNode = div(...)
val component:IO[VNode] = for {
  handler <- createHandler
} yield div( ... handler ... )
val bigComponent = for{
  component <- component
} yield div(component, purecomponent)
```
Note how pure Components don't need any `IO`. Also it is now possible to write the html-tags in `yield`.

One one hand this does not hide IO anymore. So it is more intuitive what's going on. People can transfer their Monad-Knowledge from `Future` or `Option`. On the other hand in some cases it is more cumbersome to write. I'd like to have some opinions on this. Further examples (Counter, Todo) can be seen in the ScenarioTests:
original version: https://github.com/OutWatch/outwatch/blob/master/src/test/scala/outwatch/ScenarioTestSpec.scala#L21

version of this PR: https://github.com/OutWatch/outwatch/blob/consistent-monad-interface/src/test/scala/outwatch/ScenarioTestSpec.scala#L21

When looking at code-changes, it makes sense to compare this branch against a state before referential transparency was introduced: https://github.com/OutWatch/outwatch/compare/07a0f2d2d129bd9d1990f94140c90d4e73093d8a...consistent-monad-interface

### DIFF
--- a/src/main/scala/outwatch/Sink.scala
+++ b/src/main/scala/outwatch/Sink.scala
@@ -88,14 +88,14 @@ object Sink {
     * @tparam T the type parameter of the elements
     * @return the newly created Handler.
     */
-  def createHandler[T](seeds: T*): Handler[T] = {
+  def createHandler[T](seeds: T*): IO[Handler[T]] = {
     val handler = new SubjectSink[T]
 
     if (seeds.nonEmpty) {
-      Handler(IO(ObservableSink[T](handler, handler.startWithMany(seeds: _*))))
+      IO(ObservableSink[T](handler, handler.startWithMany(seeds: _*)))
     }
     else {
-      Handler(IO(handler))
+      IO(handler)
     }
   }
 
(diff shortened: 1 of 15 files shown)


fdietze:
This would also make it trivial to provide a non-referential-transparent api by defining implicits for `IO[Handler]` to `Handler`.

mariusmuja:
I like the current referential transparent API because it largely keeps the client code unchanged. For example, while porting a small example application I only had make two small changes:

```diff
-    val inputTodo = createStringHandler()
+    createStringHandler().flatMap { inputTodo =>
...
+  }
...
-    OutWatch.render("#app", Router())
+    OutWatch.render("#app", Router()).unsafeRunSync()
``` 

With this PR the changes to the client code are more significant, requiring the change of the return type of some components from `VNode` to `IO[VNode]` and requiring for comprehensions to build the VDom  of the components that use them.

I refactored this PR further so that the client code remains mostly unchanged (just as with the current master version) and submitted it as PR #69. Interestingly, with #69 both syntax versions of the examples in ScenarioTests (the one in master and the one in this PR) work fine.

Issue #40, opened 2017-06-27
Make API referentially transparent


LukaJCB:
Kind of related to #28. Currently part of the API as well as part of the internals aren't referentially transparent. We could rewrite the `createHandler` signatures to include some kind of effect managing monad and then use for comprehensions for syntactic sugar. Here's an example:

Current Api:

    def personComponent(labelText: String, texts: Sink[String]) = {
      val firstNames = createStringHandler("")
      val lastNames = createStringHandler("")

      val fullNames = firstNames
        .combineLatestWith(lastNames)((first, last) => s"$first $last")

      div(
        label(labelText),
        input(inputString --> firstNames),
        input(inputString --> lastNames),
        button(click (fullNames) --> texts, "Submit")
      )
    }

Referentially transparent:

    def personComponent(labelText: String, texts: Sink[String]) = for {
      firstNames <- createStringHandler("")
      lastNames <- createStringHandler("")

      fullNames = firstNames
        .combineLatestWith(lastNames)((first, last) => s"$first $last")

      root <- div(
        label(labelText),
        input(inputString --> firstNames),
        input(inputString --> lastNames),
        button(click (fullNames) --> texts, "Submit")
      )

    } yield root
    

The rest of the API could probably stay the same, or atleast with minimal changes. Components without handlers should be completely backwards compatible (albeit not binary compatible). Eager to know what you guys think.

mushtaq:
The existing api looks simpler, what are the advantages of the new one? what is the return type in each case?

LukaJCB:
It is simpler, however it also has some notion of side effects currently. Whenever we call `createHandler` a new Handler gets created that is not equal to when you would call `createHandler` a second time. This breaks referential transparency and can be confusing for some. 

The new proposal would wrap these calls in some kind of Effect type (e.g. Scalaz IO), which would mean building normal VDOM trees would also be wrapped in this type. That means the return type of the `personComponent` as well as of the `div`etc. functions would be something like `VNodeBuilder`. However, I suggest renaming the current `VNode` to something like `VNodeRepresentation` and the suggested `VNodeBuilder` to `VNode`. 
This way, we can get backwards compatibility for components without handlers. :)

mushtaq:
I see that some will prefer effects to be explicitly captured, but I am finding it difficult to imagine a large code base where all calls are wrapped inside for-yield. Is it possible to keep the older api around for those who do not mind the side-effects?

LukaJCB:
[...]

PR #69, opened 2017-10-31
Another referential transparent API refactoring

mariusmuja:
I refactored the referential transparent API proposed in #68 so that it keeps the client code mostly unchanged (similarly to what the current master API does) but without wrapping the IO monad.

### DIFF
--- a/src/main/scala/outwatch/dom/helpers/DomUtils.scala
+++ b/src/main/scala/outwatch/dom/helpers/DomUtils.scala
@@ -117,7 +117,7 @@ object DomUtils {
     def toProxy(changable: (Seq[Attribute], Seq[VNode])): VNodeProxy = changable match {
       case (attributes, nodes) =>
         val updatedObj = proxy.data.withUpdatedAttributes(attributes)
-        h(proxy.sel, updatedObj, proxy.children ++ (nodes.map(_.asProxy)(breakOut):js.Array[VNodeProxy]))
+        h(proxy.sel, updatedObj, proxy.children ++ (nodes.map(_.unsafeRunSync().asProxy)(breakOut):js.Array[VNodeProxy]))
     }
 
     val subscription = changables.observable
@@ -142,17 +142,17 @@ object DomUtils {
     emitters: List[Emitter] = Nil,
     receivers: List[Receiver] = Nil,
     properties: List[Property] = Nil,
-    vNodes: List[VNode] = Nil
+    vNodes: List[VNode_] = Nil
   )
-  private[outwatch] def separateModifiers(args: Seq[VDomModifier]): SeparatedModifiers = {
+  private[outwatch] def separateModifiers(args: Seq[VDomModifier_]): SeparatedModifiers = {
     args.foldRight(SeparatedModifiers())(separatorFn)
   }
 
-  private[outwatch] def separatorFn(mod: VDomModifier, res: SeparatedModifiers): SeparatedModifiers = (mod, res) match {
+  private[outwatch] def separatorFn(mod: VDomModifier_, res: SeparatedModifiers): SeparatedModifiers = (mod, res) match {
     case (em: Emitter, sf) => sf.copy(emitters = em :: sf.emitters)
     case (rc: Receiver, sf) => sf.copy(receivers = rc :: sf.receivers)
     case (pr: Property, sf) => sf.copy(properties = pr :: sf.properties)
-    case (vn: VNode, sf) => sf.copy(vNodes = vn :: sf.vNodes)
+    case (vn: VNode_, sf) => sf.copy(vNodes = vn :: sf.vNodes)
     case (EmptyVDomModifier, sf) => sf
   }
 
@@ -187,7 +187,7 @@ object DomUtils {
     }
   }
 
-  private[outwatch] def extractChildrenAndDataObject(args: Seq[VDomModifier]): (Seq[VNode], DataObject) = {
+  private[outwatch] def extractChildrenAndDataObject(args: Seq[VDomModifier_]): (Seq[VNode_], DataObject) = {
     val SeparatedModifiers(emitters, receivers, properties, children) = separateModifiers(args)
 
     val SeparatedReceivers(childReceivers, childrenReceivers, attributeReceivers) = separateReceivers(receivers)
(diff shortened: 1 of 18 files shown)


fdietze:
Thank you for working on this! I'll have a detailed look at it as soon as possible.

mariusmuja:
@fdietze As far as I can tell it is referential transparent (what makes you think it isn't ?), for example for any `makeComponent()`:
```scala
val component = makeComponent()

val comp1 = div(component, component) 
val comp2 = div(makeComponent(), makeComponent())
```
`comp1` and `comp2` are the same.

I just added a test for this (96fc5fe), which seems to fail both for the master implementation and the one in #68, because the issue fixed by a448d1b.

src/test/scala/outwatch/ScenarioTestSpec.scala
```diff
@@ -91,6 +91,38 @@ class ScenarioTestSpec extends UnitSpec with BeforeAndAfterEach {
     document.getElementById("greeting").innerHTML shouldBe greetStart + name2
   }
 
+  "A component" should "be referential transparent" in {
```
fdietze: :+1:
This test is a very important contribution!

fdietze:
@mariusmuja Thanks for clarifying.

I see, as everything happens in IO, asProxy can never be called outside and is therefore safe.

fdietze:
I'm actually in favor of this PR now.

What is left, is a convenient way to write inline-styles.

currently one needs to write:
```scala
IO.pure(Style("color", "red"))
```

There should be a simple solution until `scala-dom-types` is integrated. Maybe just provide a function `outwatch.dom.style` like found in the tests of this PR?:
```scala
def style(property: String, value: String): IO[Style] = IO.pure(Style(property: String, value: String))
```

fdietze:
I just had the idea to drive this approach of flattening out IO even further:
```scala
  def -->(sink: Sink[E]):IO[EventEmitter[E]] =
    IO.pure(EventEmitter(eventType, sink.observer))

  def -->(sinkIO: IO[Sink[E]]):IO[EventEmitter[E]] =
    sinkIO.flatMap(-->)
```

In some cases this makes it possible to write components without for-expressions. The downside is that it looks like the old api, but unintuitively fails when combining streams. In many cases a fallback to the for-loop is needed. Any ideas?

I implemented the methods on this branch: https://github.com/mariusmuja/outwatch/compare/referential-transparent-refactor...OutWatch:referential-transparent-refactor-flatten-stream-io

mariusmuja:
@fdietze Yeah, I though about doing that a while ago (https://github.com/OutWatch/outwatch/issues/40#issuecomment-312546861), but it doesn't work as expected:

For example in this code:
```scala
val handler = createStringHandler()
div(
  button(click("clicked") --> handler),
  div(`class` := "label", child <-- handler)
)
```
the `div` doesn't get updated on `button` click, because `createStringHandler()` is now referential transparent, so it's equivalent to writing:
```scala
div(
  button(click("clicked") --> createStringHandler()),
  div(`class` := "label", child <-- createStringHandler())
)
```

fdietze:
I noticed that Sink.create() does not return an IO. Is this on purpose?

fdietze:
Here is a useful diff for reviewing: https://github.com/OutWatch/outwatch/compare/07a0f2d2d129bd9d1990f94140c90d4e73093d8a...mariusmuja:referential-transparent-refactor

fdietze:
And should `Sink` and `Observable` be imported somehow when writing `import outwatch.dom._` ? I think they should. Because else, extra imports are needed when writing component-signatures which take inputs and outputs:
```scala
def comp(input:Sink[Int], output:Observable[Int]) = ...
```

src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -202,11 +202,9 @@ object DomUtils {
   }
 
   def render(element: Element, vNode: VNode): IO[Unit] = for {
```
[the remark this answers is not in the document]
fdietze: Note that:

```scala
type VNode = IO[VNode_]
```

You just got confused by the hidden IO-Monad ;)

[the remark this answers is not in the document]
fdietze: Maybe we should not hide it and sacrifice the type-alias. What do you think?

mariusmuja:
I like the idea of having Observable and Sink imported by `outwatch.dom._`. We could do something like this:
```diff
 package object dom extends Attributes with Tags with Handlers {
 
+  type Observable[+A] = rxscalajs.Observable[A]
+  val Observable = rxscalajs.Observable
+
+  type Sink[-A] = outwatch.Sink[A]
+  val Sink = outwatch.Sink
+
   type VNode = IO[VNode_]
   type VDomModifier = IO[VDomModifier_]

```

An additional benefit is that it would ease porting Outwatch to another reactive streams library such as Monix. I prefer Monix to rxjs, I did a port here: https://github.com/mariusmuja/outwatch/tree/referential-transparent-refactor-monix

src/main/scala/outwatch/dom/package.scala
```diff
@@ -14,9 +14,9 @@ package object dom extends Attributes with Tags with Handlers {
 
   implicit def optionIsEmptyModifier(opt: Option[VDomModifier]): VDomModifier = opt getOrElse IO.pure(EmptyVDomModifier)
 
-  implicit class ioVTreeMerge(tree: IO[VTree]) {
-    def apply(args: VDomModifier*): IO[VTree] = {
-      tree.flatMap(vtree => IO.pure(VTree(vtree.nodeType, vtree.modifiers ++ args)))
+  implicit class ioVTreeMerge(vnode: VNode) {
+    def apply(args: VDomModifier*): VNode = {
+      vnode.flatMap(vnode_ => vnode_(args:_*))
```
mariusmuja: :+1:

src/test/scala/outwatch/ScenarioTestSpec.scala
```diff
@@ -153,18 +177,16 @@ class ScenarioTestSpec extends UnitSpec with BeforeAndAfterEach {
       state = adds.merge(deletes)
         .scan(Vector[String]())((state, modify) => modify(state))
         .map(_.map(n => TodoComponent(n, deleteHandler)))
-
-
-      root <- div(
-        TextFieldComponent("Todo: ", inputHandler),
+      textFieldComponent <- TextFieldComponent("Todo: ", inputHandler)
+    } yield div(
```
LukaJCB: Correct me, if I'm wrong, but here root is a `IO[IO[VNode]]` correct?

LukaJCB:
I'd just like to say thank you to all of you who put in such great work! I'm really glad to see the discussion we're having and the progress we're making!

src/test/scala/outwatch/ScenarioTestSpec.scala
mariusmuja: That is correct, it would end up flattened by the ``` vtree.flatMap``` below.

LukaJCB:
Awesome, LGTM! I'd like to bikeshed about some of the naming of internal things, but that shouldn't affect the API, so I'm going to merge this. Thank you again @mariusmuja and @fdietze  and @cornerman! :heart:

mariusmuja:
Yes, we should have better names for: `VNode_` and `VDomModifier_`.

PR #83, opened 2017-11-07
Handler comap, contramap, imap and lens

fdietze:
This is a draft for transforming handlers in the read and write direction.

1. `comap`, like `map` on `Observable`, but returns a `Handler` which can still be used for writing.
2. `contramap`, like `redirectMap` on `Observable` redirects the written values into the `Handler`, but still returns a `Handler` for reading.
3. `imap` defines an isomorphism and returns a `Handler` where reading and writing are transformed.
4. `lens` zooms into one part of the data where changes are written back to the original `Handler`.

See the tests for examples.

We could also think about accepting automatically generated lenses, Iso and other stuff from [Monocle](https://julien-truffaut.github.io/Monocle/optics.html).

For `imap` it could be useful to implement the `Invariant` type class from cats. For the `comap` and `contramap` it is sadly not possible to implement `Functor` or `Covariant`, since right now the type has to be the same. Ideas welcome!

### DIFF
--- a/src/main/scala/outwatch/Sink.scala
+++ b/src/main/scala/outwatch/Sink.scala
@@ -47,10 +47,10 @@ sealed trait Sink[-T] extends Any {
 }
 
 object Sink {
-  private final case class SubjectSink[T]() extends Subject[T](new SubjectFacade) with Sink[T] {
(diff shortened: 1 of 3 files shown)


fdietze:
I moved the relevant parts over to #91.
https://github.com/OutWatch/outwatch/pull/91/commits/d21fceb84e3335f719e84386c0d9af753921704b

PR #85, opened 2017-11-09
Refactor EmitterBuilder

fdietze:
I made `EmitterBuilder` simpler and at the same time more powerful (arbitrarily composable). This allowed me to remove many helper classes like `GenericMappedEmitterBuilder`, `StringEventEmitterBuilder` and `StringEventEmitter` (+ similar ones).

One thing I was not able to resolve yet: More possibilities made the type inference a bit worse, for example:
```scala
button(inputString(_ => number) --> stream), // _ cannot be inferred anymore,
// since arbitrary values are allowed, so the compiler does not know if you want
// to emit a function or transform the event. Therefore you have to write:
button(inputString((_:String) => number) --> stream),
```

Any ideas on the type inference? We could insert values like `inputString.as(number)`.

### DIFF
--- a/src/main/scala/outwatch/dom/Attributes.scala
+++ b/src/main/scala/outwatch/dom/Attributes.scala
@@ -39,13 +39,13 @@ trait Attributes
 /** Clipboard Events */
 trait ClipBoardEventAttributes {
   /** Fires when the user copies the content of an element. */
-  lazy val copy   = new EventEmitterBuilder[ClipboardEvent]("copy")
+  lazy val copy   = EmitterBuilder.event[ClipboardEvent]("copy")
 
   /** Fires when the user cuts the content of an element. */
-  lazy val cut    = new EventEmitterBuilder[ClipboardEvent]("cut")
+  lazy val cut    = EmitterBuilder.event[ClipboardEvent]("cut")
 
   /** Fires when the user pastes some content in an element. */
-  lazy val paste  = new EventEmitterBuilder[ClipboardEvent]("paste")
+  lazy val paste  = EmitterBuilder.event[ClipboardEvent]("paste")
 }
 
 /** Form Events that are triggered by actions inside an HTML form. However,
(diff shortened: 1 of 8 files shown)


mariusmuja:
Awesome! I added a bug fix and simplified the code further.

Regarding the type inference issue, for that particular case it could be rewritten as:

```scala
button(inputString(number) --> stream)
```

I've deprecated ```def apply[T](f: O => T)``` and replaced it with ```def map[T](f: O => T)```, which doesn't have the type inference issue, for example:
```scala
button(inputString.map(_ => number) --> stream)
```

src/main/scala/outwatch/dom/helpers/EmitterBuilder.scala
```diff
@@ -33,7 +33,7 @@ final case class EmitterBuilder[E, O] private (
 
 object EmitterBuilder {
   def apply[E <: Event](eventType: String): EmitterBuilder[E, E] = new EmitterBuilder[E, E](
-    eventType, identity, (e, o) => o.next(e.asInstanceOf[E])
+    eventType, identity, (e, o) => o.asInstanceOf[Observer[Event]].next(e)
```
fdietze: Why are you casting the `Observer` instead of the `Event`?

fdietze:
I like the deprecation of `def apply[T](f: O => T)`.

I also just inlined the trigger in `EmitterBuilder`. It was not needed anymore.

src/main/scala/outwatch/dom/helpers/EmitterBuilder.scala
mariusmuja: I was actually casting the event at one point and I was getting a weird random test failure, then I reverted to casting the observer. The failure was probably unrelated though and I believe the two ways should be equivalent (because Observer is contravariant).

mariusmuja:
I changed the `EmitterBuilder` so it's more efficient in the simple case (no map, filter or apply on the emitter) by avoiding an un-necessary `Sink.redirect` and extending AnyVal.

src/main/scala/outwatch/dom/Attributes.scala
```diff
@@ -57,60 +57,60 @@ trait FormEventAttributes {
...
+  lazy val focus        = EmitterBuilder[InputEvent]("focus")
 
   /** The input event is fired when an element gets user input. */
-  lazy val input        = new EventEmitterBuilder[InputEvent]("input")
+  lazy val input        = EmitterBuilder[InputEvent]("input")
 
 
   @deprecated("Deprecated, use 'inputChecked' instead", "0.8.0")
-  lazy val inputBool    = new BoolEventEmitterBuilder("change")
+  lazy val inputBool    = inputChecked
```
cornerman: should we take this release for removing these deprectated methods? The new release will anyhow break some user code.

cornerman:
Is the performance improvment for a non-transforming emitterbuilder really needed? The difference is that otherwise we would map an `Observable` with the `identity` function. Does this really justify adding more complexity with a transforming and non-transforming EmitterBuilder? Otherwise, LGTM.

src/main/scala/outwatch/dom/Attributes.scala
fdietze: This will be different anyways when using domtypes.

mariusmuja:
@cornerman It's more than mapping an `Observable` with the identity function, it's creating a Handler (a SubjectSink) and doing a redirect on it (`sink.takeUntil(sink.ignoreElements.defaultIfEmpty(())` ).

I don't know what the performance improvement is (some benchmarking would be needed), but don't think there's much extra complexity to have a non-transforming emitter for the default case, most of the functionality is in the common trait.

cornerman:
@mariusmuja: yeah, you are right and actually, it is not really that much code. Thanks!

PR #87, opened 2017-11-09
Incorrect child order

fdietze:
I noticed that when writing

```scala
    val vNode = div(
      "A",
      child <-- messagesA,
      "B",
      child <-- messagesB
    )
```

The order of the children is incorrect. You get `ABxx` instead. I already wrote tests, but don't have an idea for a fix yet.

The tests fail with:
```
[info] - should render child nodes in correct order *** FAILED ***
[info]   "...<span>A</span><span>[B</span><span>1]</span><span>2</span..." was not equal to "...<span>A</span><span>[1</span><span>B]</span><span>2</span..." (OutWatchDomSpec.scala:337)
[info] - should render child string-nodes in correct order *** FAILED ***
[info]   "<div>A[B1]2</div>" was not equal to "<div>A[1B]2</div>" (OutWatchDomSpec.scala:357)
```

### DIFF
--- a/src/main/scala/outwatch/dom/helpers/DomUtils.scala
+++ b/src/main/scala/outwatch/dom/helpers/DomUtils.scala
@@ -2,47 +2,21 @@ package outwatch.dom.helpers
 
 import cats.effect.IO
 import org.scalajs.dom._
+import org.scalajs.dom.html
 import outwatch.dom._
 import rxscalajs.Observable
+import rxscalajs.subjects.BehaviorSubject
 import rxscalajs.subscription.Subscription
 import snabbdom._
-import collection.breakOut
 
+import collection.breakOut
 import scala.scalajs.js
 import scala.scalajs.js.JSConverters._
 
 
 object DomUtils {
 
-  private case class Changeables(attributeStreamReceivers: Seq[AttributeStreamReceiver],
-                                 childrenStreamReceivers: Seq[ChildrenStreamReceiver],
-                                 childStreamReceivers: Seq[ChildStreamReceiver]) {
-    lazy val observable: Observable[(Seq[Attribute], Seq[VNode])] = {
-      val childReceivers: Observable[Seq[VNode]] = Observable.combineLatest(
-        childStreamReceivers.map(_.childStream)
-      )
-
-      val childrenReceivers = childrenStreamReceivers.lastOption.map(_.childrenStream)
-
-      // only use last encountered observable per attribute
-      val attributeReceivers: Observable[Seq[Attribute]] = Observable.combineLatest(
-        attributeStreamReceivers
-          .groupBy(_.attribute)
-          .values
-          .map(_.last.attributeStream)(breakOut)
-      )
-
-      val allChildReceivers = childrenReceivers.getOrElse(childReceivers)
-
-      attributeReceivers.combineLatest(allChildReceivers)
-    }
-
-    lazy val nonEmpty: Boolean = {
-      attributeStreamReceivers.nonEmpty || childrenStreamReceivers.nonEmpty || childStreamReceivers.nonEmpty
-    }
-  }
-
-  private def createDataObject(changeables: Changeables,
+  private def createDataObject(changeables: SeparatedReceivers,
                                properties: Seq[Property],
                                eventHandlers: js.Dictionary[js.Function1[Event, Unit]]): DataObject = {
 
@@ -66,7 +40,7 @@ object DomUtils {
     DataObject.create(attrs, props, style, handlers, insertHook, deleteHook, updateHook, key)
   }
 
-  private def createReceiverDataObject(changeables: Changeables,
+  private def createReceiverDataObject(changeables: SeparatedReceivers,
                                        properties: Seq[Property],
                                        eventHandlers: js.Dictionary[js.Function1[Event, Unit]]) = {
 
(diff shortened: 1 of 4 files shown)


mariusmuja:
I think with the current implementation there's supposed to be just one child receiver / vnode. In most cases there is a workaround:
```scala
    val vNode = div(
      "A",
      span(child <-- messagesA),
      "B",
      span(child <-- messagesB)
    )
```

fdietze:
@mariusmuja Thank you for fixing this.

As a side note:
I just rebased. This keeps the git-history clean and therefore easier to work with. I suggest you to do the same when updating your branches.
When working on the same branch, just do `git pull --rebase` instead of `git pull` to avoid unnecessary merge commits.
When you need updates from another branch, do `git rebase <other branch>`. This makes the most sense for keeping pull-requests up-to-date with the current master-branch.

With these practices merging PRs become a fast-forward merge.

fdietze:
While reviewing, I figured out, that we have a similar problem with children:
https://github.com/OutWatch/outwatch/pull/87/commits/7ecf790fadbc24e64d8515e907bd1221be5affe5

mariusmuja:
With the latest change it's possible to mix child streams and children streams. Also, it's possible to have multiple children streams.

However, although possible, I'm not sure it's a good practice to do so, when there is a children stream it should be the only one in the node.

fdietze:
I agree that this is not good practice, but I think it should be either forbidden or behave in a predictable way.

I just tried squashing and then rebasing this branch on the current master. But I have problems resolving the conflics, since I didn't write the code... @mariusmuja could you please do that?

mariusmuja:
@fdietze Yes, I'll do it later today.

src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -201,12 +199,22 @@ object DomUtils {
...
   }
 
+
+  private final case class ChildrenNodes(
+    children: List[VNode_] = Nil,
+    hasStreams: Boolean = false
+  )
+  private def extractChildren(nodes: Seq[ChildVNode]): ChildrenNodes = nodes.foldRight(ChildrenNodes()) {
+    case (vn: VNode_, cn) => cn.copy(children = vn :: cn.children)
+    case (_, cn) => cn.copy(hasStreams = true)
```
cornerman: This seems a bit dangerous, as we are only assuming that the type of `_` must be a stream and then set `hasStreams` to true. If someone would refactor the inheritance of `ChildVNode`, this usage not be catched by the compiler.

src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -172,16 +146,40 @@ object DomUtils {
...
+    childNodes: List[ChildVNode] = Nil,
+    hasNodeStreams: Boolean = false,
+    attributeStreamReceivers: List[AttributeStreamReceiver] = Nil
+  ) {
+
+    lazy val observable: Observable[(Seq[Attribute], Seq[VNode])] = {
+
+      val childStreamReceivers = if (hasNodeStreams) {
+        childNodes.foldRight(Observable.of(List.empty[VNode])) {
+          case (vn: VNode_, obs) => obs.combineLatestWith(BehaviorSubject(IO.pure(vn)))((nodes, n) => n :: nodes)
```
cornerman: I do not get why we need to include `vn` into the observable here. Isn't it handled separately when we collect the `children` in `ChildrenNodes`?

mariusmuja: No, those children get displayed just initially. After a `child`/`children` stream element gets emitted, the  parent node gets patched and those children would be removed during the node update if they were not included here. 

In master, the children get preserved during a patch operation, but in the incorrect order. In this PR, they are included in the node change observable in the correct order.

src/main/scala/outwatch/dom/helpers/DomUtils.scala
mariusmuja: Yes, that's correct. I've updated the PR.

src/main/scala/outwatch/dom/helpers/DomUtils.scala
mariusmuja: I realized that in order for the patch operation to be efficient, we need to generate snabbdom `key`s (if not already present) for all the original children that get included in the node observable.

src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -178,18 +174,45 @@ object DomUtils {
...
+    case (vn: VNode_, cn) => cn.copy(children = vn :: cn.children)
+    case (_: ChildStreamReceiver, cn) => cn.copy(hasStreams = true)
+    case (_: ChildrenStreamReceiver, cn) => cn.copy(hasStreams = true)
+  }
+
+  private[outwatch] def ensureVNodeKey(vn: VNode_): VNode_ = {
+    val withKey: VNode_ = vn match {
+      case vtree: VTree =>
+        val modifiers = vtree.modifiers
+        val hasKey = modifiers.exists(m => m.unsafeRunSync().isInstanceOf[Key])
```
cornerman: Is it right to `run` the `IO` here and then wrap the modifier itself again as an `IO`?

mariusmuja: `ensureVNodeKey` only gets called inside a snabbdom `patch`, so I think it's ok, it doesn't affect the referential transparency of the Outwatch API. 

Also `ensureVNodeKey` gets called from `extractChildrenAndDataObject` which gets called from `Vtree.asProxy` which already does `unsafeRunSync` on modifiers.

Note that it's not doing `.unsafeRunSync` and then wrapping inside `IO` again, it's just adding a `key` modifier to the existing modifiers.

fdietze:
@mariusmuja Thank you very much for fixing this one!

PR #90, opened 2017-11-11
make ObservableSink public

### LINKED ISSUE #79: Make combinator Sink.ObservableSink public
It is very useful to be able to combine a sink and an observable into a new handler, so please make [`ObservableSink`]( https://github.com/OutWatch/outwatch/blob/master/src/main/scala/outwatch/Sink.scala#L53) public.

cornerman:
fixes #79

### DIFF
--- a/src/main/scala/outwatch/Sink.scala
+++ b/src/main/scala/outwatch/Sink.scala
@@ -50,10 +50,6 @@ object Sink {
   private final case class SubjectSink[T]() extends Subject[T](new SubjectFacade) with Sink[T] {
     override private[outwatch] def observer = this
   }
-  private final case class ObservableSink[T]
-  (oldSink: Sink[T], stream: Observable[T]) extends Observable[T](stream.inner) with Sink[T] {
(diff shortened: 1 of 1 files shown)


mariusmuja:
Having `ObservableSink`'s constructor public could be the source of unexpected bugs if the `ObservableSink` is not constructed with an `Observable` that gets it's elements from the `Sink`.

@cornerman Can you give more some examples with where you would need this? Maybe there's another solution.

cornerman:
@mariusmuja I think most of my needs would be catered by https://github.com/OutWatch/outwatch/pull/83. But in general for extending the handler and enabling to define such functions yourself, it might make sense to expose some mapping function. How about we make it more explicit and introduce something like `Sink.mapHandler`:

```scala
  def mapHandler[T, R](handler: Handler[T])(mapSink: Sink[T] => Sink[R], mapObservable: Observable[T] => Observable[R]): Handler[R] =
    ObservableSink(mapSink(handler), mapObservable(handler))
```

PR #95, opened 2017-11-21
build ChildStreamReceiver only from VNodes

cornerman:
The `ChildStreamReceiverBuilder` accepts `Observable[Any]` which removes type safety and introduces runtime suprises. By accident I can easily write Options, Futures, or other undesired things into the value. This feels like the java-inherited `+` and `any2stringadd` in scala.

I have changed the Builder to only accept types that have an implicit conversion defined. Currently, only strings are automatically converted to `VNode`. So some tests fail as they require `Int` to work, too. Should we define an implicit for primitive types or require an explicit `toString` or what do you expect?

### DIFF
--- a/src/main/scala/outwatch/VNodeRender.scala
+++ b/src/main/scala/outwatch/VNodeRender.scala
@@ -1,0 +1,24 @@
+package outwatch
+
+import cats.effect.IO
+import outwatch.dom.{StringNode, VNode}
+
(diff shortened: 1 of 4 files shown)


mariusmuja:
How about using a typeclass for rendering `VNode`s for primitive types?

cornerman:
Looks good to me!

PR #99, opened 2017-11-22
Handle boolean properties

mariusmuja:
Same as for attributes, boolean properties must be handled correctly. For example, in the Chrome console:
```
> elem
<input type=​"checkbox" name=​"collab_privs" class=​"js-collab-option" checked>​
> elem.checked
true
> elem.checked = "false"
"false"
> elem.checked
true
> elem.checked = false
false
> elem.checked
false
```

### DIFF
--- a/src/main/scala/outwatch/dom/VDomModifier.scala
+++ b/src/main/scala/outwatch/dom/VDomModifier.scala
@@ -25,8 +25,9 @@ object Attribute {
 
 
 final case class Attr(title: String, value: String | Boolean) extends Attribute
-final case class Prop(title: String, value: String) extends Attribute
+final case class Prop(title: String, value: String | Boolean) extends Attribute
 final case class Style(title: String, value: String) extends Attribute
+
 final case class Key(value: String) extends Property
 
 sealed trait Hook extends Property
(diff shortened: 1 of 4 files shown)


fdietze:
Isn't this already handled by #62 ? @cornerman ?

cornerman:
Good catch! This is also not handled in the scala-dom-types branch. Maybe fix it there?

mariusmuja:
Yes, fixing it in #62.

cornerman:
Actually, I think we do not need to handle this. You are just experiencing type conversion in javascript, the `truthy` value `"false"` is interpreted as `true`. You can set it to `""` or `0` and it is false. Now `checked` is just a property, which is the current value of the radio/checkbox. And the attribute `checked` which you set in your html element corresponds to the property `defaultChecked`. Initially checked attribute and property are the same but they are not [reflected](https://github.com/raquo/scala-dom-types#reflected-attributes).

For snabbdom, boolean handling only exists for `attributes` and not for `properties` (see https://github.com/snabbdom/snabbdom#the-props-module - this readme is a bit outdated with regards to boolean attributes, check the release text for that part: https://github.com/snabbdom/snabbdom/releases/tag/v0.7.0).

mariusmuja:
I was actually using `checked` as a prop on my branch at it wasn't working properly until I fixed this. I'll test if the issue persists with #62 and if it does, I'll add a test for it.

cornerman:
Thanks, please add a test there and check whether you need the commit for boolean properties (best with [a reflected attribute like e.g. `disabled`](https://github.com/raquo/scala-dom-types/blob/master/shared/src/main/scala/com/raquo/domtypes/generic/defs/reflectedAttrs/ReflectedAttrs.scala#L207). You can hopefully remove the commit :)

mariusmuja:
I verified that we need to handler boolean props on #62. I pushed a fix and a test case.

PR #101, opened 2017-11-24
Link source maps to raw.githubusercontent.com

fdietze:
I saw this interesting snippet in [build.sbt](https://github.com/lihaoyi/scala.rx/blob/master/build.sbt#L57) and thought , that it might be useful for debugging. But I wasn't able to verify that its working.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,6 +27,12 @@ npmDependencies in Compile ++= Seq(
   "snabbdom" -> "0.7.0"
 )
 
+scalacOptions += {
+  val local = baseDirectory.value.toURI
+  val remote = s"https://raw.githubusercontent.com/OutWatch/outwatch/${git.gitHeadCommit.value.get}/"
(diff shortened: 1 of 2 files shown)


mariusmuja:
Nice! But I think it should be:
```diff
-  val a = baseDirectory.value.toURI.toString.replaceFirst("[^/]+/?$", "")
+  val a = baseDirectory.value.toURI
```

fdietze:
I cannot find any relevant url in the source-map files. I think this is for source-maps from scalaJS itself, but bundler does not adopt it yet.

mariusmuja:
The urls are in the `.sjsir` files and I verified that the bundler includes them in the `.map` files.

fdietze:
I now included the current git hash in the URL, which should always load the correct source file.
I tested this successfully with chromium, which now shows the source file when debugging.

Try it with this dependency:

```scala
"com.github.fdietze" % "outwatch" % "3f5cd8e"
```

I suggest that other libraries use this approach as well.

PR #102, opened 2017-11-25
HandlerFactories should be functions

cornerman:


LukaJCB:
What's wrong with the current encoding?

cornerman:
Why should the handler factory methods be lazy vals? Both should work, as the implicit class is instantiated every time I call one of the properties (`Handler.inputEvents`, `Handler.mouseEvents`, ...). So the vals would not persist, but there is also no advantage over using def, which is cheaper.

LukaJCB:
Aaah yes that makes perfect sense! +1

Issue #105, opened 2017-11-26
Problem with updating user-changeable properties


cornerman:
Before scala-dom-types, we used the `value` attribute and manually synced this with property `value` (valueSyncHook). This was specific for the `value` property and still left problems with e.g. `checked`. Now, we use the property `value` directly and not the attribute. This generally fixes the problem, but now we are running in into an issue with updating user-changeable properties in snabbdom: https://github.com/snabbdom/snabbdom/issues/345.

mariusmuja:
React.js has the notion of `controlled`/`un-controlled` components, the difference being the source of truth about the state of the component (the DOM or the component).

For example translating the example from snabbdom/snabbdom#345 to Outwatch it could be something like:

```scala
import outwatch.dom.{Handler, Attributes => A, Tags => T}

def component(): VNode = {
  Handler.create[String].flatMap { strHandler =>
    T.div(
      T.input(A.`type` := "text", A.value := "test", A.value <-- strHandler),
      T.button(A.onClick("test") --> strHandler, "Set")
    )
  }
}
```

Here, pressing the "Set" button after changing the input text doesn't do anything because the the new prop'a value is the same as the old prop.

We can turn this into a "controlled component" (in react.js terminology), by sync-ing the DOM state to the component, which fixes the issue:

```scala
def component(): VNode = {
  Handler.create[String].flatMap { strHandler =>
    T.div(
      T.input(A.`type` := "text", A.value := "test", A.value <-- strHandler, A.onInputString --> strHandler),
      T.button(A.onClick("test") --> strHandler, "Set")
    )
  }
}
```

For a nicer syntax, we could have a `valueControlled` VDomModifier builder:

```scala
object valueControlled {
  def <-->(handler: Handler[String]): VDomModifier = Seq(
    A.value <-- handler, A.onInputString --> handler
  )
}

def component(): VNode = {
  Handler.create[String].flatMap { strHandler =>
    T.div(
      T.input(A.`type` := "text", A.value := "test", valueControlled <--> strHandler),
      T.button(A.onClick("test") --> strHandler, "Set")
    )
  }
}

```

fdietze:
I'm also in favor of not expecting too much knowledge about virtual dom inner workings from the user. It should be simple and just work.

I researched a bit and never found any benchmark or evidence that reading dom properties may be slow. Does anyone know something more about that? Benchmarks would be nice.

When looking at constructs like

```scala
A.value <-- strHandler, A.onInputString --> strHandler
[...]

PR #108, opened 2017-11-29
custom snabbdom props module

cornerman:
As discussed in #105, this adds a custom snabbdom props module, which compares against the actual element property.

**Please do not merge**, as I think, the generated code for scala-js is not as performant as a pure javascript implementation. So, we should implement this module in js and somehow include it with scalajs-bundler/webpack.

What do you think?

### DIFF
--- a/src/main/resources/snabbdom-custom-props.js
+++ b/src/main/resources/snabbdom-custom-props.js
@@ -1,0 +1,33 @@
+"use strict";
+
+// original: https://github.com/snabbdom/snabbdom/blob/master/src/modules/props.ts
+
+function updateProps(oldVnode, vnode) {
+    var key, cur, old, elm = vnode.elm, oldProps = oldVnode.data.props, props = vnode.data.props;
+
+    if (!oldProps && !props) return;
+    if (oldProps === props) return;
+    oldProps = oldProps || {};
+    props = props || {};
+
+    for (key in oldProps) {
+        if (!props[key]) {
+            delete elm[key];
+        }
+    }
+
+    for (key in props) {
+        cur = props[key];
+        old = elm[key];
+        if (old !== cur) {
+            elm[key] = cur;
+        }
+    }
+}
+
+module.exports = {
+    default: {
+        create: updateProps,
+        update: updateProps
+    }
+};
--- a/src/main/scala/snabbdom/h.scala
+++ b/src/main/scala/snabbdom/h.scala
@@ -136,7 +136,7 @@ object patch {
     SnabbdomClass.default,
     SnabbdomEventListeners.default,
     SnabbdomAttributes.default,
-    SnabbdomProps.default,
+    SnabbdomCustomProps.default,
     SnabbdomStyle.default
   ))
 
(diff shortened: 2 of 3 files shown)


cornerman:
here is an example of how to have custom javascript resources: https://github.com/scalacenter/scalajs-bundler/tree/master/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/js-resources

mariusmuja:
After reading snabbdom/snabbdom#345, I'd like to do some benchmarking on how this affects the performance for patching larger `VNode`s that contain multiple props.

fdietze:
@cornerman which properties are used by outwatch?

cornerman:
only [these five props](https://github.com/raquo/scala-dom-types/blob/master/shared/src/main/scala/com/raquo/domtypes/generic/defs/props/Props.scala): value, checled, selected, textContent, indeterminate.

If we wanted we could also use the prop module for the [reflected attributes](https://github.com/raquo/scala-dom-types/blob/master/shared/src/main/scala/com/raquo/domtypes/generic/defs/reflectedAttrs/ReflectedAttrs.scala), but there we currently use attributes.

fdietze:
The current alternatives to the custom snabbdom-module are:
* The old valueSyncHook (also reads from dom, so probably worse performance)
* Via Handlers (`<-->`) from #105 (updates the dom on every keystroke)

Before taking the time to benchmark this approach, are there any other (theoretically) faster alternatives?

cornerman:
exactly, the `ValueSyncHook` did actually the same before for the `value` property:

```scala
  private val valueSyncHook: (VNodeProxy, VNodeProxy) => Unit = (_, node) => {
    node.elm.foreach { elm =>
      val input = elm.asInstanceOf[HTMLInputElement]
      if (input.value != input.getAttribute("value")) {
        input.value = input.getAttribute("value")
      }
    }
  }
```

mariusmuja:
Agreed, all alternatives for keeping the DOM in sync seem to have similar trade-offs. Except that with the current snabbdom props, the DOM `value` only gets updated if it changed and the `-->` gets updates via `Events`, not from reading the DOM.

So, I don't know what the performance of the `<-->` alternative is, as for `value`s it doesn't update the DOM on every keystroke, it reads the DOM on every keystroke event and skips updating because it didn't change)... something to benchmark.

Just occurred to me about reflected attributes... what is the performance of updating an attribute vs updating a prop (with the current snabbdom props module). Would the performance be better to use props instead for reflected attributes ( since most reflected attributes don't get changed by another agent)

fdietze:
> So, I don't know what the performance of the <--> alternative is, as for values it doesn't update the DOM on every keystroke, it reads the DOM on every keystroke event and skips updating because it didn't change)... something to benchmark.

I don't understand what you are trying to say here. Most keystrokes on an input field change the value. So it only skips in the rare cases of keys which don't change it (e.g. arrow keys). Am I overlooking something?

mariusmuja:
> I don't understand what you are trying to say here. Most keystrokes on an input field change the value. So it only skips in the rare cases of keys which don't change it (e.g. arrow keys). Am I overlooking something?

I was thinking of the current props implementation that handles `value` as a special case and only writes the DOM prop it if the value changed compared to the actual DOM value. So, for something like this:

```scala
input(tpe := "text", value <-- strHandler, onInputString --> strHandler),
```
there is no DOM prop write on each keystoke, only a DOM prop read.

... but that's exactly what the proposed props module does too. And since outwatch is only updating 5 props though the props module and the rest are reflected attributes, I'm not so worried any more about the performance impact (I was forgetting earlier that we're using reflected attributes for all other props and I was worried about the performance impact of reading the DOM when updating each of them).

cornerman:
> Just occurred to me about reflected attributes... what is the performance of updating an attribute vs updating a prop (with the current snabbdom props module). Would the performance be better to use props instead for reflected attributes ( since most reflected attributes don't get changed by another agent)

This is still a valid question, but I do not know of any benchmarks and how syncing between props and attributes actually behaves.

Issue #93, opened 2017-11-14
InputEvent assumes incorrect type for target


cornerman:
Currently, we have [`InputEvent`](https://github.com/OutWatch/outwatch/blob/master/src/main/scala/outwatch/dom/helpers/InputEvent.scala#L10) for representing all [form events](https://github.com/OutWatch/outwatch/blob/master/src/main/scala/outwatch/dom/Attributes.scala#L55). But InputEvent assumes the wrong type of the `eventTarget` and casts to `HTMLInputElement`. Even events like `onInput` can trigger on elements like `<input>`, `<textarea>`, `<select>`, or `<... contenteditable>`. The other events have different possible element targets.

We try to tackle this in `scala-dom-types` and for most events we can only assume `html.Element` for typing the `eventTarget`. It is better than type `EventTarget`, but the problem is, you still do not have access to the `.value`, `.valueAsNumber`, `.checked` properties (we currently have that).

Three options for not having to write `.asInstanceOf[HtmlInputElement].value`:

1) Define helper methods like `targetValue`, `targetValueAsNumber`, `targetChecked` on the event. Something like [this](https://github.com/raquo/scala-dom-types/blob/master/js/src/main/scala/com/raquo/domtypes/jsdom/defs/package.scala#L16):
```scala
  implicit class InputEventWithValue(val ev: InputEvent) extends AnyVal {
    import dom.html

    def targetValue: String = ev.target match {
      case input: html.Input => input.value
      case textarea: html.TextArea => textarea.value
      case select: html.Select => select.value
      case elem => elem.textContent // for contenteditable
    }

    def targetValuesAsNumber: Int = ...
    def targetChecked: Boolean = ...
  }
```

I kind of like this option, as it is short and will work for any element that could be the target. But has to make certain assumptions, especially for implementing something like `targetValueAsNumber`, `targetChecked`.

2) Define helper methods `targetT` as a shortcut for `target.asInstanceOf`:
```scala
  implicit class EventWithTypedTarget(val ev: dom.Event) extends AnyVal {
    def targetT[T <: dom.EventTarget]: T = ev.target.asInstanceOf[T]
  }
```

More flexible, as is `asInstanceOf`, but not typesafe. If you ever come to change the embedding element, it will fail.

3) Define helper methods like `targetInput`, `targetTextarea`, `targetSelect`. Similar like the one before, but saving you from importing the type parameter.
```scala
[...]

PR #110, opened 2017-12-02
Typed currentTarget for events

cornerman:
This introduces the trait [`TypedCurrentTargetEvent`](https://github.com/cornerman/outwatch/blob/current-target/src/main/scala/outwatch/dom/TypedCurrentTargetEvent.scala#L9), which allows to have a concrete type for the `currentTarget` of events. The `currentTarget` of an event always corresponds to the element, on which the event listener was registered. We have this information at compile time when constructing the dom elements:

```scala
input(
  tpe := "text",
  onChange.map(_.currentTarget.valueAsNumber) --> intSink
)
``` 

For this to compile, we need to know the type of the outer element `input` while mapping the `onChange` event EmitterBuilder. As already discussed in #93, type inference does not help here, therefore I have introduced a `macro` for injecting the element type to have the correct type for the `currentTarget`.

1) We introduce a [`TagContext`](https://github.com/cornerman/outwatch/blob/current-target/src/main/scala/outwatch/dom/TagContext.scala#L8) to capture the type of the Element and let the apply method of `VTree` accept a function of `TagContext[Element] => Seq[VDomModifier]`. Furthermore, we create events as `UnassignedEvent` first and then [implicitly convert](https://github.com/cornerman/outwatch/blob/current-target/src/main/scala/outwatch/dom/Events.scala#L24) it to an `EmitterBuilder` with more information about the Element. We can now write this and pick up the TagContext and thereby the outer type implicitly:
```scala
input(implicit tagContext => Seq(
  tpe := "text",
  onChange.map(_.currentTarget.valueAsNumber) --> intSink
))
```

2) Introduce [`VTree.apply` as macro](https://github.com/cornerman/outwatch/blob/current-target/src/main/scala/outwatch/dom/VDomModifier.scala#L75) to get rid of the boilerplate. But there is one problem: as of now, macros need their arguments to typecheck before the macro is even called (`valueAsNumber` needs to typecheck on an unknown outer element). So we need a [globally available TagContext](https://github.com/cornerman/outwatch/blob/current-target/src/main/scala/outwatch/dom/TagContext.scala#L13) with a somehow [powerful Element type](https://github.com/cornerman/outwatch/blob/current-target/src/main/scala/outwatch/dom/TagContext.scala#L26) that convinces the compiler to typecheck all thinkable usages of `currentTarget`. Don't worry, this type is only used at compile time and will be [replaced by the macro](https://github.com/cornerman/outwatch/blob/current-target/src/main/scala/outwatch/dom/Macros.scala), which injects the correct element type which then passes the typechecker again. 

This allows us to write the construct as desired.

One downside is the usage of events outside of a `VTree.apply`, but we inform the user about this via `compileTimeOnly` error messages. You there need to provide an implicit for the TagContext or explicitly bind the event to an element type:
```scala
def modifiers(implicit ctx: TagContext[html.Input]) = Seq(
  id := "1",
  onClick.map(_.currentTarget.value) --> sink
)

let modifiers = Seq(
  id := "1",
  onClick.onElement[html.Input].map(_.currentTarget.value) --> sink
)
```

From what I have tested, there is no huge hit on compile time. Intellij copes with it okayish. It will autocomplete for `DummyElement` and cannot highlight the wrong type, but the compiler will still tell you the right thing and there are no false compile errors in the editor.

Further changes:
* `VTree[Element]` generic on the corresponding `dom.Element`
* `VTree` public and remove `apply` from `VNode`
* Deprecate `onInputString`, `onInputNumber`, `onInputChecked`


cornerman:
What we are still missing in here is typesafety with regards to the element type of `VDomModifier`.So, for example, I could do this: `div(onInput.onElement[html.Input].map(_.currentTarget.value) --> sink)`. This is wrong, as the currentTarget of the onInput subscription on the div element can never be of type `Input`. To solve this, we would need to make the div.apply method only accept `VDomModifier[html.Div]`. This would also be handy for restricting certain properties to specific tags, see https://github.com/raquo/scala-dom-types/issues/13.

cornerman:
I have added some fixes and [a new convenience typeclass for getting the `string|number|boolean` value from an element](https://github.com/cornerman/outwatch/blob/current-target/src/main/scala/outwatch/dom/TagWith.scala). This allows to write something like this:

```scala
      for {
        stringStream <- Handler.create[String]
        doubleStream <- Handler.create[Double]
        boolStream <- Handler.create[Boolean]
        elem <- input(
            onSearch.target.string --> stringStream,
            onClick.currentTarget.string --> stringStream,
            onSearch.target.number --> doubleStream,
            onClick.currentTarget.number --> doubleStream,
            onSearch.target.boolean --> boolStream,
            onClick.currentTarget.boolean --> boolStream
          )
        )
      } yield elem
```

In this case, writing `onClick.currentTarget.string` is just a shorthand for `onClick.map(_.currentTarget.value)`. if you switch the outer element to `textArea` only string related methods are defined on currentTarget, but with e.g. `div` none of these will compile. Additionally, `TagWithString` allows to write generic methods for components, for example:

```scala
    def stringModifier[Elem <: Element : TagContext : TagWithString]: VDomModifier =
      onChange.currentTarget.string --> stringHandler
```

And then use it like:

```scala
    div(
       input(stringModifier), textArea(stringModifier)
    )
```

mariusmuja:
I think having typesafe `target` and `currentTarget` is great, but I'm not convinced that the added complexity that this PR introduces is worth it. I particularly dislike the use of macros to build the VDom.

cornerman:
For `target`, this comes from domtypes, so we do not need to make any additional effort for using this type.

For the `currentTarget`, it is different. I agree that a macro is not the most beautiful solution but it allows to keep the normal syntax, while additionally getting typeinformation about the embedding type. If you all feel strongly about macros, we need to introduce a different concept for getting the currentTarget type, an alternative syntax without macros could be:

```scala
input(
  tpe := "text",
  _.typed(onChange).map(_.currentTarget.valueAsNumber) --> intSink // _.typed() picks up the type from the parent
  onChange.map((ev: TypedCurrentTargetEvent[html.Input]) => ev.currentTarget.valueAsNumber) --> intSink // providing the type explicitly also works
  // onChange.map(_.currentTarget.valueAsNumber) --> intSink // will not compile, as no type information about currentTarget is provided
)
```

I feel that the complexity introduced by the macro is still maintainable and it is nice to have a clean syntax for events without additional clutter.

cornerman:
@LukaJCB How do you think about macros?

cornerman:
> I particularly dislike the use of macros to build the VDom.

Can you kindly elaborate why that is and what you think about the mentioned alternatives?

mariusmuja:
> Can you kindly elaborate why that is and what you think about the mentioned alternatives?

The reason is that macros are an experimental feature, they have poor IDE support, they increase the maintenance burden of the library and also make the library harder to understand by users (I often like to browse through the code of the libraries I'm using to figure out how various things work, and macros often make that much harder).
So, I generally avoid macros unless there's a very compelling reason to use them. For this particular case I'm not sure the gain is worth the added complexity.
Regarding the alternatives, I haven't had a chance to look over #119 yet, I'll try to do that soon.

cornerman:
superseded by #119

PR #114, opened 2017-12-03
move TagsCompat to Tags trait

cornerman:


cornerman:
Let's try to keep a clean git history with no ff merge commits, so rebase and merge is a better choice for small changes :)

PR #115, opened 2017-12-06
Child/children node streams optimisations 

mariusmuja:
This PR fixes an issue that caused static nodes to be destroyed/re-inserted when the first `child`/`children` element was emitted.

It also ensures that no `Seq.empty[VNode]` is emitted at the beginning of the `children` stream  when a single `children` stream is present in the node (the common case).

### DIFF
--- a/src/main/scala/outwatch/dom/VDomModifier.scala
+++ b/src/main/scala/outwatch/dom/VDomModifier.scala
@@ -4,7 +4,7 @@ import cats.effect.IO
 import org.scalajs.dom._
 import outwatch.dom.helpers.DomUtils
 import rxscalajs.Observer
-import snabbdom.{DataObject, VNodeProxy, h}
+import snabbdom.{DataObject, VNodeProxy, hFunction}
 
 import scala.scalajs.js
 import collection.breakOut
@@ -25,11 +25,21 @@ object Attribute {
 }
 
 final case class Attr(title: String, value: Attr.Value) extends Attribute
-object Attr { type Value = DataObject.AttrValue }
+object Attr {
+  type Value = DataObject.AttrValue
+}
+
 final case class Prop(title: String, value: Prop.Value) extends Attribute
-object Prop { type Value = DataObject.PropValue }
+object Prop {
+  type Value = DataObject.PropValue
+}
+
 final case class Style(title: String, value: String) extends Attribute
-final case class Key(value: String) extends Property
+
+final case class Key(value: Key.Value) extends Property
+object Key {
+  type Value = DataObject.KeyValue
+}
 
 sealed trait Hook extends Property
 final case class InsertHook(observer: Observer[Element]) extends Hook
@@ -53,8 +63,8 @@ sealed trait VNode_ extends ChildVNode {
 }
 
 //TODO: extends AnyVal
-private[outwatch] case class StringNode(string: String) extends VNode_ {
-  val asProxy: VNodeProxy = VNodeProxy.fromString(string)
+private[outwatch] final case class StringNode(string: String) extends VNode_ {
+  override val asProxy: VNodeProxy = VNodeProxy.fromString(string)
 }
 
 // TODO: instead of Seq[VDomModifier] use Vector or JSArray?
@@ -63,19 +73,19 @@ private[outwatch] case class StringNode(string: String) extends VNode_ {
 private[outwatch] final case class VTree(nodeType: String,
                        modifiers: Seq[VDomModifier]) extends VNode_ {
 
-  def asProxy = {
+  override def apply(args: VDomModifier*) = IO.pure(VTree(nodeType, modifiers ++ args))
+
+  override def asProxy = {
     val modifiers_ = modifiers.map(_.unsafeRunSync())
     val (children, attributeObject) = DomUtils.extractChildrenAndDataObject(modifiers_)
     //TODO: use .sequence instead of unsafeRunSync?
     // import cats.instances.list._
     // import cats.syntax.traverse._
     // for { childProxies <- children.map(_.value).sequence }
-    // yield h(nodeType, attributeObject, childProxies.map(_.apsProxy)(breakOut))
+    // yield hFunction(nodeType, attributeObject, childProxies.map(_.apsProxy)(breakOut))
     val childProxies: js.Array[VNodeProxy] = children.map(_.asProxy)(breakOut)
-    h(nodeType, attributeObject, childProxies)
+    hFunction(nodeType, attributeObject, childProxies)
   }
-
-  override def apply(args: VDomModifier*) = IO.pure(VTree(nodeType, modifiers ++ args))
 }
 
 
--- a/src/main/scala/outwatch/dom/helpers/DomUtils.scala
+++ b/src/main/scala/outwatch/dom/helpers/DomUtils.scala
@@ -50,7 +50,7 @@ object DomUtils {
     val insertHook = createInsertHook(changeables, subscriptionRef, insert)
     val deleteHook = createDestroyHook(subscriptionRef, destroy)
     val updateHook = createUpdateHook(update)
-    val key = keys.lastOption.map(_.value).getOrElse(changeables.hashCode.toString)
+    val key = keys.lastOption.fold[Key.Value](changeables.hashCode)(_.value)
 
     DataObject.create(attrs, props, style, eventHandlers, insertHook, deleteHook, updateHook, key)
   }
@@ -71,7 +71,7 @@ object DomUtils {
 
     def toProxy(changable: (Seq[Attribute], Seq[VNode])): VNodeProxy = {
       val (attributes, nodes) = changable
-      h(
+      hFunction(
         proxy.sel,
         proxy.data.withUpdatedAttributes(attributes),
         if (nodes.isEmpty) proxy.children else nodes.map(_.unsafeRunSync().asProxy)(breakOut): js.Array[VNodeProxy]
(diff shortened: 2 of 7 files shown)


src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -46,15 +56,15 @@ final case class ChildStreamReceiver(childStream: Observable[VNode]) extends Chi
 final case class ChildrenStreamReceiver(childrenStream: Observable[Seq[VNode]]) extends ChildVNode
 
 sealed trait VNode_ extends ChildVNode {
-  // TODO: have apply() only on VTree?
```
cornerman: We need to keep this TODO!

src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -63,19 +73,27 @@ private[outwatch] case class StringNode(string: String) extends VNode_ {
...
+    // yield hFunction(nodeType, attributeObject, childProxies.map(_.apsProxy)(breakOut))
     val childProxies: js.Array[VNodeProxy] = children.map(_.asProxy)(breakOut)
-    h(nodeType, attributeObject, childProxies)
+    hFunction(nodeType, attributeObject, childProxies)
   }
 
-  override def apply(args: VDomModifier*) = IO.pure(VTree(nodeType, modifiers ++ args))
+  // ensure a key is present in the VTree modifiers
+  // used to ensure efficient Snabbdom patch operation in the presence of children streams
+  private[outwatch] def ensureKey: VTree = {
```
cornerman: Does this method really belong here? Maybe better to leave it where it is used.

src/main/scala/snabbdom/hFunction.scala
```diff
@@ -57,13 +61,14 @@ trait DataObject extends js.Object {
...
   val hook: Hooks
-  val key: js.UndefOr[String | Int]
+  val key: js.UndefOr[KeyValue]
 }
 
 object DataObject {
 
   type PropValue = Any
   type AttrValue = String | Boolean
+  type KeyValue = String | Int
```
cornerman: Where is this definition coming from?

src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -4,7 +4,7 @@ import cats.effect.IO
 import org.scalajs.dom._
 import outwatch.dom.helpers.DomUtils
 import rxscalajs.Observer
-import snabbdom.{DataObject, VNodeProxy, h}
+import snabbdom.{DataObject, VNodeProxy, hFunction}
```
fdietze: Why this rename? Instead of importing `hFunction` one could just write `snabbdom.h( ... )` everywhere instead.

src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -46,15 +56,15 @@ final case class ChildStreamReceiver(childStream: Observable[VNode]) extends Chi
...
+  def apply(args: VDomModifier*): VNode
   // TODO: rename asProxy to asSnabbdom?
   def asProxy: VNodeProxy
 }
 
 //TODO: extends AnyVal
-private[outwatch] case class StringNode(string: String) extends VNode_ {
-  val asProxy: VNodeProxy = VNodeProxy.fromString(string)
+private[outwatch] final case class StringNode(string: String) extends VNode_ {
+  override def apply(args: VDomModifier*): VNode = ???
```
fdietze: This should stay as default implementation in `VNode_` and be handled via the TODO instead.

src/main/scala/snabbdom/hFunction.scala
fdietze: https://github.com/snabbdom/snabbdom#key--string--number

src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -50,7 +50,7 @@ object DomUtils {
     val insertHook = createInsertHook(changeables, subscriptionRef, insert)
     val deleteHook = createDestroyHook(subscriptionRef, destroy)
     val updateHook = createUpdateHook(update)
-    val key = keys.lastOption.map(_.value).getOrElse(changeables.hashCode.toString)
+    val key = keys.lastOption.fold[Key.Value](changeables.hashCode)(_.value)
```
fdietze: good catch!

src/main/scala/snabbdom/hFunction.scala
fdietze: And should we use `Double` instead of `Int` here? (because of `Number`) Javascript only knows `Double` anyways, so this should be a noop even when using `Int`-keys.

fdietze:
Good that you noticed, fixed and provided tests to prevent these issues in the future. When the minor changes in the comments are addressed, I will approve. :+1:

src/main/scala/outwatch/dom/VDomModifier.scala
mariusmuja: I renamed it because I think `h` is a pretty poor name for it (makes sense in JS where it's used a lot directly and it should be short, but that's not the case here). Also, the trait was already called `hFunction` and having a trait and an object with the same name is common in Scala.js facades.

It's still possible to have `import snabbdom.{hFunction => h}` if somebody really likes the `h` name.

src/main/scala/outwatch/dom/VDomModifier.scala
mariusmuja: I updated the PR and removed the `apply()` from `VNode_`.

src/main/scala/outwatch/dom/package.scala
```diff
@@ -25,8 +25,8 @@ package object dom extends Attributes with Tags with HandlerFactories {
 
   implicit def compositeModifier(modifiers: Seq[VDomModifier]): VDomModifier = IO.pure(CompositeVDomModifier(modifiers))
 
-  implicit class ioVTreeMerge(vnode: VNode) {
-    def apply(args: VDomModifier*): VNode = {
+  implicit class ioVTreeMerge(vnode: IO[VTree]) {
```
cornerman: Should we introduce `type VTree = IO[VTree_]`, as we did for VNode and VDomModifier? We will need to use this type for the apply method.

mariusmuja:
Rebased and squashed commits.

src/main/scala/outwatch/dom/package.scala
mariusmuja: I updated the PR so that the signature of `ioVTreeMerge` remains unchanged. If we can avoid it, I think it's better to not expose `VTree`, the fewer types the users have to deal with the better.

src/main/scala/outwatch/dom/package.scala
```diff
@@ -26,8 +26,11 @@ package object dom extends Attributes with Tags with HandlerFactories {
   implicit def compositeModifier(modifiers: Seq[VDomModifier]): VDomModifier = IO.pure(CompositeVDomModifier(modifiers))
 
   implicit class ioVTreeMerge(vnode: VNode) {
-    def apply(args: VDomModifier*): VNode = {
-      vnode.flatMap(vnode_ => vnode_(args:_*))
+    def apply(args: VDomModifier*): VNode= {
+      vnode.flatMap {
+        case vtree: VTree => vtree(args: _*)
+        case sn: StringNode => IO.pure(sn)
```
cornerman: That is not what I meant and I think we should not silently drop these modifiers. Instead only VTree should have an apply method.

It can easily happen to construct something like this:

```
    def myNode: VNode = {
      "hans"
    }

    div(myNode(contentEditable := "hans"))
```

cornerman: Additionally, I still think I would be beneficial to let `VTree` know about its dom.Element type, so that this becomes `VTree[Element]`, so you can reason about the type of that node. This would also require exposing this type.

cornerman: Maybe better to leave this as it was in this PR and tackle it again in another one?

mariusmuja: > It can easily happen to construct something like this and then the modifier is just gone:

Hmm... how would a modifier get applied to a string node? Currently, that results in runtime `NotImplementedError` because of the `???`.

cornerman: Exactly, it should not be possible to apply a modifier to a string node. And this should be enforced by the typesystem, so that apply is not defined on `StringNode`. Of course, a `NotImplementedError` is wrong, too :)

mariusmuja: > Maybe better to leave this as it was in this PR and tackle it again in another one?

Yeah, makes sense to limit the scope of this PR, I'll revert the change.

cornerman: I think a string node is normally only used within a VNode (e.g. span("hello world")). Maybe, we could make `StringNode extends VDomModifier_` and deal with it internally? Then, there would only be `VTree` extending `VNode_`.

mariusmuja: > I think a string node is normally only used within a VNode (e.g. span("hello world")). Maybe, we could make StringNode extends VDomModifier_ and deal with it internally? Then, there would only be VTree extending VNode_.

Yes, I though about that and I like that solution, but it breaks usages like this:

```scala
val stringStream: Observable[String]
div(child <-- stringStream)
```

Possible solutions would be to have a conversion from `StringNode` to `VNode` only for the stream case or to disallow this use it and provide a `div(textChild <-- stringStream)`.

An additional argument to make `StringNode` a `VDomModifier` is that currently at the Snabbdom level something like `span("foo")` ends up translated to `{'sel': 'span', children: [{'text': 'foo'}]}` instead of `{'sel': 'span', 'text': 'foo'}`.

cornerman: > Possible solutions would be to have a conversion from StringNode to VNode only for the stream case or to disallow this use it and provide a div(textChild <-- stringStream).

I like the first option more, because it keeps the api for the user. We could just overload `def <--(stream: Observable[String])`.

cornerman:
LGTM. Thank you for working on this!

PR #116, opened 2017-12-06
Hooks optimisations and refactorings

mariusmuja:
Currently, even if there are no hook sinks present on a `VNode`, there are empty insert/update/create hooks created for all the `VNode`s. This PR leaves the hooks as `undefined` in case there are no hook sinks present.

The PR also adds support for the `prepatch`/`postpatch` hooks. I needed `postpatch` for frameworks such [Material Design Lite](https://getmdl.io/) and [Materialize](http://materializecss.com/) that require dynamic components to be "upgraded" after they are inserted in the DOM (using the `update` hooks did not work, the "upgrading" has to happen after the snabbdom patch is completed).

### DIFF
--- a/src/main/scala/outwatch/dom/VDomModifier.scala
+++ b/src/main/scala/outwatch/dom/VDomModifier.scala
@@ -41,10 +41,17 @@ object Key {
   type Value = DataObject.KeyValue
 }
 
-sealed trait Hook extends Property
-final case class InsertHook(observer: Observer[Element]) extends Hook
-final case class DestroyHook(observer: Observer[Element]) extends Hook
-final case class UpdateHook(observer: Observer[(Element, Element)]) extends Hook
+sealed trait Hook[T] extends Property {
+  def observer: Observer[T]
+}
+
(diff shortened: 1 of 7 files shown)


src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -32,9 +32,25 @@ final case class Style(title: String, value: String) extends Attribute
 final case class Key(value: String) extends Property
 
 sealed trait Hook extends Property
-final case class InsertHook(observer: Observer[Element]) extends Hook
-final case class DestroyHook(observer: Observer[Element]) extends Hook
-final case class UpdateHook(observer: Observer[(Element, Element)]) extends Hook
+
+sealed trait HookSingle extends Hook {
```
cornerman: Can we make these traits generic? So we have `def observer: Observer[T]`. We can later pattern match on the value when mapping.

fdietze: If there is a way to make this trait generic and combine the functions `createHookSingle` etc. , great. But if that turns out to be impractical, I'm fine with this solution.

PR #117, opened 2017-12-07
Refactorings to the VDomModifier hierarchy.

### LINKED ISSUE #113: EmptyAttribute for BooleanSelector attributes
For the `BooleanSelector` (`?=`), we currently write an attribute with key `hidden` into the dom: https://github.com/OutWatch/outwatch/blob/master/src/main/scala/outwatch/util/SyntaxSugar.scala#L11. We should have an `EmptyAttribute` like `EmptyVDomModififer`.

mariusmuja:
I did some refactorings to the VDomModifier hierarchy so that `StringNode` is no longer a `VNode_`, instead is a `VDomModifier_`, so that it was possible to remove the `apply(args: VDomModifier*)` method from `VNode_` which was failing with a runtime exception if somebody tried to apply modifiers to a `StringNode`.

As a consequence, `VNode_` is replaced by `VTree` and I introduced a `StaticVNode` that can be either a `StringVNode` or a `VTree`, but the `StringVNode` is only used in those cases a where a `StringModifier` would not work, for example when using something like:

```scala
val stringStream: Observable[String]
div(child <-- stringStream)
```
This PR also fixes the way strings are converted to Snabbdom in most cases, for example currently  something like `span("foo")` ends up translated to `{'sel': 'span', children: [{'text': 'foo'}]}` instead of `{'sel': 'span', 'text': 'foo'}`. This is fixed, except for those cases when a `StringVNode` is required, like in the example above.

I re-ordered the traits/classes in VDomModifier.scala in breath-first order so that similar classes (Attributes, Properties, ...) are next to each other.

This PR is on top of #115.

Fixes #113.


src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -9,13 +9,31 @@ import snabbdom.{DataObject, VNodeProxy, hFunction}
 import scala.scalajs.js
 import collection.breakOut
 
-sealed trait VDomModifier_ extends Any
+sealed trait Modifier extends Any
```
cornerman: Lets keep this PR focused and not rename `VDomModifier`. Otherwise, we have more conflicts than needed on other branches.

src/main/scala/outwatch/dom/DomTypes.scala
```diff
@@ -14,11 +14,11 @@ import org.scalajs.dom
...
 
 private[outwatch] object DomTypesBuilder {
-  type GenericVNode[T] = VNode_
+  type GenericVNode[T] = VTree
 
-  trait VNodeBuilder extends TagBuilder[GenericVNode, VNode_] {
+  trait VNodeBuilder extends TagBuilder[GenericVNode, VTree] {
     // we can ignore information about void tags here, because snabbdom handles this automatically for us based on the tagname.
-    override def tag[Ref <: VNode_](tagName: String, void: Boolean): VNode_ = VTree(tagName, Seq.empty)
+    protected override def tag[Ref <: VTree](tagName: String, void: Boolean): VTree = VTree(tagName, Seq.empty)
```
cornerman: When we make this protected, we need to introduce a `def tag` in OutwatchAttributes for as we did for `attr`, `prop`, `style`. Then we should also add a testcase for custom tags.

src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -9,13 +9,31 @@ import snabbdom.{DataObject, VNodeProxy, hFunction}
...
 
-case class Emitter(eventType: String, trigger: Event => Unit) extends VDomModifier_
+// Modifiers
 
-sealed trait Property extends VDomModifier_
+sealed trait Property extends Modifier
+
+final case class Emitter(eventType: String, trigger: Event => Unit) extends Modifier
+
+private[outwatch] final case class AttributeStreamReceiver(attribute: String, attributeStream: Observable[Attribute]) extends Modifier
```
cornerman: Why are these private now?

fdietze:
I think we should establish a practice of submitting separate rename and code-move PRs. This makes it a lot easier to review and resolve conflicts.

src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -24,6 +42,17 @@ object Attribute {
...
 }
 
+
+sealed trait Hook[T] extends Property {
+  def observer: Observer[T]
+}
+
+// Attributes
+
+case object EmptyAttribute extends Attribute {
```
cornerman: can somehow not have an empty string as a title here?

src/main/scala/outwatch/dom/VDomModifier.scala
mariusmuja: It seemed a good time to rename it because `VNode_` got removed and together with `VDomModifier_` they had names I didn't like. But it makes sense regarding conflict with other branches.

src/main/scala/outwatch/dom/VDomModifier.scala
mariusmuja: Because it should only be used from within the outwatch package. Is that not the case?

src/main/scala/outwatch/dom/VDomModifier.scala
mariusmuja: No, `EmptyAttribute`s are skipped and not passed to Snabbdom.

mariusmuja:
I reverted the `VDomModifier_` rename and added a `def tag()` for custom tag creation.

cornerman:
I have added a trait `TitledAttribute` to distinguish between EmptyAttribute (without title) and normal attributes (with title).

Issue #23, opened 2017-03-07
ClassSet


PerWiklander:
This makes it easier to include a bunch of classes depending on boolean values (taken from scalars-react).

```scala
div(
  classSet(
    "message"           -> true,
    "message-active"    -> true,
    "message-important" -> props.isImportant,
    "message-read"      -> props.isRead),
  props.message)

// Or for convenience, put all constants in the first arg:
div(
  classSet1(
    "message message-active",
    "message-important" -> props.isImportant,
    "message-read"      -> props.isRead),
  props.message)
```
Do we have some Observable magic to take into account here? Would we want an `Observable[Boolean]` to be able to do the toggling?

LukaJCB:
I'm not sure what exactly can be toggled here?

PerWiklander:
The classes that have a Boolean true in the tuple will be included in the class string. So you could have a class attribute value of "disabled large button" but you only want "disabled" in there depending on a Boolean.

 /Per

fdietze:
Snabbdom supports this feature already: https://github.com/snabbdom/snabbdom#the-class-module

PerWiklander:
So we need some glue code that translates a `Seq[(String, Boolean)]` into a JS literal? How about `Observable[Boolean]`?

fdietze:
First the DataObject (https://github.com/OutWatch/outwatch/blob/master/src/main/scala/snabbdom/h.scala#L52) needs an additional field `class:js.Dictionary[Boolean]`.

Supporting `Boolean` and `Observable[Boolean]` on the usage side sounds good to me.

cornerman:
We will add a `toggle` method for attributes (including classes), which should improve this significantly (see #445). With this, I do not see a need for supporting the classset as proposed.

Therefore, I am closing this issue. Feel free to reply or reopen if there is more to discuss.

PR #118, opened 2017-12-08
Add support for accumulated attributes (className in particular)

mariusmuja:
This PR adds support for attributes that can be accumulated within the same `VNode`, in particular for accumulating `className`s within a `VNode`.

I often find this useful when I'm using some component and I want to pass in some additional `className` modifier to the component without overriding the class(es) that the component already uses.

For example:

```scala

def component(modifiers: VDomModifier*): VNode

val disabled = classAccum := "disabled-class"

if (someCondition) component(modifiers: _*) else component( (disabled +: modifiers) : _*)
```

### DIFF
--- a/src/main/scala/outwatch/dom/DomTypes.scala
+++ b/src/main/scala/outwatch/dom/DomTypes.scala
@@ -27,7 +27,7 @@ private[outwatch] object DomTypesBuilder {
   }
 
   object CodecBuilder {
-    type Attribute[T, _] = AttributeBuilder[T]
+    type Attribute[T, _] = ValueBuilder[T, Attr]
     type Property[T, _] = PropertyBuilder[T]
 
     def encodeAttribute[V](codec: Codec[V, String]): V => Attr.Value = codec match {
@@ -95,13 +95,18 @@ trait ReflectedAttrs
   extends reflectedAttrs.ReflectedAttrs[CodecBuilder.Attribute]
   with ReflectedAttrBuilder[CodecBuilder.Attribute] {
 
+  // super.className.accum(" ") would have been nicer, but we can't do super.className on a lazy val
+  override lazy val className = new AccumAttributeBuilder[String]("class",
+    stringReflectedAttr(attrKey = "class", propKey = "className"),
+    _ + " " + _
+  )
+
   override protected def reflectedAttr[V, DomPropV](
     attrKey: String,
     propKey: String,
     attrCodec: Codec[V, String],
     propCodec: Codec[V, DomPropV]
-  ): AttributeBuilder[V] =
-    new AttributeBuilder(attrKey, CodecBuilder.encodeAttribute(attrCodec))
+  ) = new AttributeBuilder(attrKey, CodecBuilder.encodeAttribute(attrCodec))
     //or: new PropertyBuilder(propKey, propCodec.encode)
 }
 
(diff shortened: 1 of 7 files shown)


fdietze:
#23 comes to my mind. What about using the snabbdom [class](https://github.com/snabbdom/snabbdom#the-class-module) module instead? Or is your use-case different?

cornerman:
I am not totally convinced about the use-case here, which is very specific for `class`. Do we really need to introduce this new concept of `AccumAttr`, which would only work for `Attr`?

mariusmuja:
> #23 comes to my mind. What about using the snabbdom class module instead? Or is your use-case different?

Using snabbdom's class module would work, but it seems to use the `classList` prop which has worse browser support (https://github.com/snabbdom/snabbdom#using-classes-in-svg-elements). It  would be less flexible, one could for example accumulate multiple Boolean signals into a boolean `disabled` attribute for example.

mariusmuja:
> I am not totally convinced about the use-case here, which is very specific for class. Do we really need to introduce this new concept of AccumAttr, which would only work for Attr?

Hmm, well I have a use case for it, and apparently others too (#23, Snabbdom's class module).

cornerman:
> Using snabbdom's class module would work, but it seems to use the classList prop which has worse browser support (https://github.com/snabbdom/snabbdom#using-classes-in-svg-elements).

Is this really a problem? This looks quite green: https://caniuse.com/#feat=classlist

Another question: how would and should `classAccum` behave with `class`?
```scala
div(classAccum := "hans", classAccum := "peter", class := "anders")
```

mariusmuja:
> Is this really a problem? This looks quite green: https://caniuse.com/#feat=classlist

Yes, it seems to be well supported (except IE 11), but still not as good as `className`. My comment was related to the suggestion in the Snabbdom's  readme to use the `attr` module instead of the `class` module for SVG elements.

> Another question: how would and should classAccum behave with class?

> div(classAccum := "hans", classAccum := "peter", `class` := "anders")

`classAccum` accumulates all attributes before it. In case there is a of a `class` following like in the example above that would override all classes. This could result in unexpected behaviour for some users (but arguably so is one class modifier being overridden by another class modifier following it). 

My preference would be for `class` to behave like `classAccum` by default, so there's no need for `classAccum` (this is also how the `class` modifier in the `scalajs-react` VDom behaves and I "discovered" the need for `classAccum` while porting some code from `scalajs-react` to `outwatch`).

cornerman:
Ok, I am convinced and let's make `classAccum`, the default for `class`.

mariusmuja:
Rebased on top of #123, made `classAccum` the default behaviour for `className` and removed `classAccum`.

src/main/scala/outwatch/dom/DomTypes.scala
```diff
@@ -92,6 +92,8 @@ trait ReflectedAttrs
   extends reflectedAttrs.ReflectedAttrs[CodecBuilder.Attribute]
   with ReflectedAttrBuilder[CodecBuilder.Attribute] {
 
+  override lazy val className: AccumAttributeBuilder[String] = new AccumAttributeBuilder("class", _.toString, _ + " " + _)
```
cornerman: We also need to overwrite cls from domtypes!

mariusmuja: You are correct, I was just noticing that. I declared `AttributesCompat` and `TagsCompat` self types to `Attributes` and `Tags`, which should take care of that I think.

However, there's still a compilation issue on 2.11, do you have any idea why? Seems to only happen when importing `className` form the `dom` package object, using `Attributes.className` works fine.

cornerman: Hmm, I have no idea. It is really weird working in the dom package, because it has all these properties automagically. But it should have worked as it was before? We can clean it up with #124.

PR #119, opened 2017-12-10
Macro-free typed currentTarget for events

cornerman:
Another take on https://github.com/OutWatch/outwatch/pull/110 without macros.

Most of it stays the same and VTree-apply accepts a function `TagContext[Elem] => VDomModifier`, where TagContext inherits all event properties from domtypes. So you can use it like this:
```scala
input(
  onClick("something") --> stringStream, // old behaviour works as before

  _.onClick.map(_.currentTarget.value) --> stringStream,
  _.onClick.value --> stringStream,
  onClick.value --> stringStream, // does not compile as onClick does not know about embedding tag

  onSearch.target.map(_.target.value) --> stringStream,
  onSearch.target.value --> stringStream,
  onClick.target.value --> stringStream, // does not compile as onClick has no typed target like onSearch
  onClick.targetAs[html.TextArea].value --> stringStream // shortcut for asInstanceOf if target is needed instead of currentTarget
  onClick.map(_.targetAs[html.TextArea].value) --> stringStream // shortcut for asInstanceOf if target is needed instead of currentTarget
)
```

We can also write generic methods with these types:
```scala
def modifier(ctx: TagContext[html.Input]): VDomModifier = Seq(
  ctx.onDrag.value --> stringHandler,
  ctx.onDrag.map(_.currentTarget.value) --> stringHandler
)
def stringModifier[Elem <: Element : TagWithString](ctx: TagContext[Elem]): VDomModifier = Seq(
  ctx.onDrag.value --> stringHandler,
  ctx.onDrag.map(_.currentTarget.value) --> stringHandler
)

div(
    textArea(modifier, stringModifier), input(modifier, stringModifier)
)
```


cornerman:
Can we assume that `target` is typed the same as `currentTarget` on `input|button|textarea` elements, because these elements cannot have children and therefore an event cannot bubble up but must be from the element itself?

cornerman:
rebased on top of #123

mariusmuja:
First of all great work on this PR, achieving the type safety without macros!

However, I'm still not sure that the added complexity is worth it for a typed `currentTarget`. `currentTarget` will have the same type as the node it's used in, so I think using a cast is fine. With this PR I find that the API around the `target`/`currentTarget` is getting un-necessarily complicated.

Currently we have:
```scala
input(
  onClick.map(_.currentTarget.asInstanceOf[html.Input].value) --> stringStream
)
```
There's not confusion what the tag type is and what's going on (cast of `currentTarget` because the event happens inside `input`).

with this PR we have:
```scala
input( 
// copy-paste from above
  _.onClick.map(_.currentTarget.value) --> stringStream,
  _.onClick.value --> stringStream,
  onClick.value --> stringStream, // does not compile as onClick does not know about embedding tag

  onSearch.target.map(_.target.value) --> stringStream,
  onSearch.target.value --> stringStream,
  onClick.target.value --> stringStream, // does not compile as onClick has no typed target like onSearch
  onClick.targetAs[html.TextArea].value --> stringStream // shortcut for asInstanceOf if target is needed instead of currentTarget
  onClick.map(_.targetAs[html.TextArea].value) --> stringStream // shortcut for asInstanceOf if target is needed instead of currentTarget)
```
Here I have no idea what the target type is (what's `target`, what's `currentTarget`) without going into the library code to figure out what implicit conversions get applied.

Also, because of the implicit conversions for every VDomModifier, the VDOM generation code seems to be getting 3x-ish larger (in fastOpt, not sure in fullOpt).

I've been torn about this PR, because I like the typesafety, but after going through the code and trying to use the API, I realized that it makes using the library harder for me, not easier and I think that's more important.

cornerman:
> There's not confusion what the tag type is and what's going on (cast of currentTarget because the event happens inside input).

While that is true, you will need to make sure that both the outer tag and the type in `asInstanceOf` stay in sync. How should we go about `onInputString`, `onInputNumber` and `onInputChecked`? They are currently hiding this assumption and just silently do an asInstanceOf.

> Here I have no idea what the target type is (what's target, what's currentTarget) without going into the library code to figure out what implicit conversions get applied.

Hmm, confusion is of course not the intention. You always know what type `currentTarget` is because it is by definition the surrounding tag, in this case `input`. For target, you know as much as MDN knows and this information is already there coming from domtypes.

I can understand that this might introduce too many variations. But we could also leave out all the shortcuts and just have `input( onClick.map(_.currentTarget.value) )`. This should not be confusing as it is just following the definition of `currentTarget` on `Event`.

> Also, because of the implicit conversions for every VDomModifier, the VDOM generation code seems to be getting 3x-ish larger (in fastOpt, not sure in fullOpt).

This indeed is a problem! I am also not happy to have this kind of conversion, but I do not really see another way to provide that type information and write it both ways. I will think about it.

mariusmuja:
> While that is true, you will need to make sure that both the outer tag and the type in asInstanceOf stay in sync. How should we go about onInputString, onInputNumber and onInputChecked? They are currently hiding this assumption and just silently do an asInstanceOf.

Those are indeed problematic. We could make them take a type argument and use a the `TagWith...` typeclasses, something like:

```scala
def onInputString[E <: dom.html.Element](implicit tag: TagWithString[E])  = Events.onInput.map(e => tag.value(e.currentTarget.asInstanceOf[E]))
```

So then they would be used as:
```scala
input( onInputString[html.Input] --> stringStream)
```

cornerman:
Why not generalize for other events and have these methods on the EmitterBuilder? I think this is not only useful for `onChange|onInput`

```
input(
  onClick.stringValue[html.Input] --> ???,
  onChange.numberValue[html.Input] --> ???
)
```

mariusmuja:
Yes, that looks good. Then we could alias commonly used sources, such as `inputString`, `inputNumber`, `inputChecked` as:

```scala
  lazy val inputChecked = onChange.checked[dom.html.Input]
  lazy val inputNumber  = onInput.numberValue[dom.html.Input]
  lazy val inputString  = onInput.stringValue[dom.html.Input]
```

I'm suggesting to use `inputChecked` (not `onInputChecked`), to indicate that it's a  `checked` event on the `input` tag.

cornerman:
funnily `inputString` would also work on `TextArea` or any other element that has a `value`, because javascript cannot check the type when casting and accessing value will work. So, maybe we should make the naming more about the property we are accessing?

mariusmuja:
So what you're suggesting is:
```scala
  lazy val checked = onChange.checked[dom.html.Input]
  lazy val valueAsNumber  = onInput.numberValue[dom.html.Input]
  lazy val value  = onInput.stringValue[dom.html.Input]
```
?
That would work too.

fdietze:
I'd rather keep the `on`-prefix for emitters.

cornerman:
> So what you're suggesting is:

Exactly, but we can extend it for different kinds of events and define these methods on the EmitterBuilder - letting the user chose the right event property. I would suggest to take parts of what I have done in this PR and combine it with the unsafe cast approach:

```scala
input(
  // these would all work on currentTarget which will always be the input element
  onInput.value --> string,
  onClick.checked --> bool,
  onKeyDown.valueAsNumber --> number
)
```

We can additionally offer the variant with providing a type (e.g. `onInput.stringValue[html.Input]`) or we just leave it at this for now.

Then there are use cases, where you might want to choose `target` over `currentTarget`. You have a `div` element with `input` children and want to collect the changed values. Then it would be useful to do the same with target:

```scala
div(
  onChange.target.value --> string,
  input(tpe := "text"), input(tpe := "text")
)
```

mariusmuja:
I played around with this here: https://github.com/mariusmuja/outwatch/compare/monix_port...mariusmuja:target_ops

I got it to work the following way: 
 * `currentTarget` uses `html.Input` by default as the underlying event target:
```scala
input(
  onInput.value --> string,
  onClick.checked --> bool,
  onKeyDown.valueAsNumber --> number
)
```
 *  it's possible to force a different event target for `currentTarget`, but only if the `TagWith...` typeclasses are implemented for it (so only `html.TextArea`, `html.Select` in case of `.value`).
 ```scala
textArea(
  onChange.value[html.TextArea] --> string,
)
```
* for `target`, the type from `TypedTargetEvent` is used by default
```scala
input(
  onSearch.target.value --> stringStream,
  onSearch.target.valueAsNumber --> doubleStream,
  onSearch.target.checked --> boolStream,
)
```
* in case the type of `TypedTargetEvent` is not specific enough it's possible to force a more specific type:
```scala
input(
  //onClick.target.value --> stringStream,   // does not compile because onClick is TypedTargetEvent[dom.Element], not specific enough
  onClick.target[html.Input].value --> stringStream,
  onClick.target[html.Input].valueAsNumber --> doubleStream,
  onChange.target[html.Input].checked --> boolStream,
)
```

cornerman:
closing for now. this is partially covered by #139. I still think, we will eventually need something like this.

PR #120, opened 2017-12-10
Refactor and fix DomUtils

cornerman:
On the basis of https://github.com/OutWatch/outwatch/pull/117, I have refactored DomUtils to be more typesafe and fold only once over the Modifiers.

Additionally, I have added a test and a fix for two issues we currently have:
1. The key we `ensure` was not unique
2. The `IO[VDomModifier_]` was run multiple times

PR #122, opened 2017-12-10
Bug with running IO[VTree] multiple times

cornerman:
Sadly, this is not a solution, but just a failing test case.


cornerman:
@mariusmuja Any ideas what is happening?

cornerman:
This is now fixed in #120 :)

Issue #73, opened 2017-11-01
Move snabbdom facade to separate project


fdietze:
This in combination with scala-dom-types would make the codebase of outwach much simpler.

cornerman:
Files in the `snabbdom` package should not depend on outwatch. Eventually, this would make it possible to move snabbdom to a different repo.

Currently, it depends on `outwatch.dom` in these places:
- https://github.com/OutWatch/outwatch/blob/master/src/main/scala/snabbdom/h.scala#L115
- https://github.com/OutWatch/outwatch/blob/master/src/main/scala/snabbdom/VDomProxy.scala

cornerman:
As is, we use a fork of snabbdom and additionally made some amendments to the facade. Therefore, this does not really apply anymore.

fdietze:
It could still make sense to have the facade in a separate sbt sub-project.

cornerman:
Sure, we can do that :)

PR #123, opened 2017-12-11
Modifier separation refactorings

### LINKED ISSUE #70: Provide DomUtils.renderReplace()
Currently `render(element, vnode)` creates an additional element inside the given one. It would be nice to be able to patch the given element.

This should also be added to the `OutWatch` object:
https://github.com/OutWatch/outwatch/blob/b455d4196ed4677883ee93463482a38467e9e4c3/src/main/scala/outwatch/dom/OutWatch.scala#L8

### LINKED ISSUE #74: Allow OutWatch.render(document.body, ...)
Currently `OutWatch.render` only accepts a string and passes that in `document.querySelector`:
https://github.com/OutWatch/outwatch/blob/b455d4196ed4677883ee93463482a38467e9e4c3/src/main/scala/outwatch/dom/OutWatch.scala#L10

It would be nice to overload `render` to also accept dom-nodes.

mariusmuja:
On top of #120, I did some more refactoring to the modifier separation, so that there's a single fold over all the `VDomModifier_` elements in a `VTree`. I've tried to keep the hierarchy created during the modifier separation mirror closely the `VDomModifier_` hierarchy, so that the code is easy to follow.

Also, I've moved the logic that creates the Snabbdom data structures to a different file (Snabbdom.scala) and removed `outwatch` dependencies from the `snabbdom` package, makes fixing #73 easy.

Moved `DomUtils.render` to `OutWatch` object, fixes #70, fixes #74.


cornerman:
Very nice work on the separation!

cornerman:
Are we ready to merge this?

mariusmuja:
I was waiting for a second review before merging it. 
@fdietze @LukaJCB

PR #125, opened 2017-12-11
add EmitterBuilder collect method

cornerman:


mariusmuja:
This is in #123 already.

cornerman:
interesting :)

mariusmuja:
Yeah, I had it in the monix branch for a while and I also did the modifier separation on the monix branch first. When I merged from the monix branch I kept it (should have been a different PR).

PR #127, opened 2017-12-12
rename LifecycleHooks with on prefix

cornerman:


fdietze:
I like this, because it makes the API more consistent. 👍

PR #129, opened 2017-12-13
Fix cleanup of jsdom environment

### LINKED ISSUE #59: Nondeterministic test fail in DomEventSpec
```
[info] - should emit and receive events correctly *** FAILED ***
[info]   null was not equal to "" (DomEventSpec.scala:37)
```
https://github.com/OutWatch/outwatch/blob/104e907ef756edf76e5b3af415f30f4407ac605c/src/test/scala/outwatch/DomEventSpec.scala#L37

Travis: https://travis-ci.org/OutWatch/outwatch/jobs/280875080#L1092

fdietze:
The problem was that dom nodes of previous tests leaked into later tests. This was problematic if the previous test used the same id, in this case `id="btn"`. The nondeterminism came from the nondeterministic test execution order. Some test classes didn't clean up the dom which was problematic, if executed in a certain order.

fixes #59

### DIFF
--- a/src/test/scala/outwatch/DomEventSpec.scala
+++ b/src/test/scala/outwatch/DomEventSpec.scala
@@ -1,25 +1,11 @@
 package outwatch
 
 import org.scalajs.dom._
-import org.scalatest.BeforeAndAfterEach
-import org.scalatest.prop.PropertyChecks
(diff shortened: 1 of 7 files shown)


mariusmuja:
Thanks for fixing this!

Issue #96, opened 2017-11-21
Make Outwatch FRP-backend agnostic


fdietze:
Different projects already use different FRP/Streaming/State management libraries like `rxjs`, `Monix`, `Xstream`, `scala.rx` or `Diode` to fit their needs.

I think it is time to make Outwatch FRP-backend agnostic to not force the user to use `rxjs`. It would be the first scala-js frontend library to do this (while at the same time extracting the dom-types: #62). If other libraries follow to adopt this flexibility, the eco-system of frontend-libraries becomes more compatible and lowers the barrier to develop better ones.

So how could we provide a solid interface which represents the data-capabilities every UI-Component needs?

My first Idea would be to define two separate type-classes for reading and writing data:

```scala
@typeclass trait Source[+T] {
     def subscribe(onNext: T => Unit):Unit
}

@typeclass trait Sink[-T] {
     def next(value: T): Unit
}
```
(I probably didn't think of every use-case yet, but you get the idea)

The question is if it is even possible to extract an interface which is flexible enough and is still enough to build all existing UI-Libraries upon.

For example: Outwatch internally already uses operators on Observables like `map` and `combineLatest`. What are other libraries using?

What do you think?

@raquo @mariusmuja @LukaJCB @cornerman
(If we think this is possible, we should also invite other frontend library developers to jump in and define a great and flexible interface.)

raquo:
Yep, I'm using `map` / `mapTo` / `filter` / `merge` and maybe some others internally in Laminar. And `unsubscribe` of course.

---

So basically you're proposing to create implementations of such standardized Sink / Source traits in terms of end user's favorite FP library. Then Outwatch or other libraries will be able to use the standardized methods internally, and the end user will be able to use the full gamut of their FP library, right?

This could probably work. Selecting the right set and amount of methods for the standardized traits would be one of the hard parts.

---

Another concern is the small details in subscriptions semantics. For example, xstream.js, my library of choice for now, unsubscribes from streams asynchronously, i.e. the subscription will be gone only after `setTimeout(..., 0)`. Other libraries could be doing it synchronously. That might or might not be a problem, I don't know.

Same goes for how different libraries handle various FRP glitches. Hard for me to say if that would be a problem.

---

[...]

PR #130, opened 2017-12-13
Monix port

mariusmuja:
This is a port of Outwatch to the [Monix](https://monix.io/) library.

### DIFF
--- a/src/main/scala/outwatch/Pipe.scala
+++ b/src/main/scala/outwatch/Pipe.scala
@@ -2,7 +2,8 @@ package outwatch
 
 import cats.effect.IO
 import outwatch.Sink.{ObservableSink, SubjectSink}
-import rxscalajs.Observable
+import outwatch.dom.Observable
+
 
 object Pipe {
   private[outwatch] def apply[I, O](sink: Sink[I], source: Observable[O]): Pipe[I, O] =
@@ -18,7 +19,7 @@ object Pipe {
     */
   def create[T](seeds: T*): IO[Pipe[T, T]] = create[T].map { pipe =>
     if (seeds.nonEmpty) {
-      pipe.transformSource(_.startWithMany(seeds: _*))
+      pipe.transformSource(_.startWith(seeds))
     }
     else {
       pipe
--- a/src/main/scala/outwatch/Pipe.scala
+++ b/src/main/scala/outwatch/Pipe.scala
@@ -1,6 +1,7 @@
 package outwatch
 
 import cats.effect.IO
+import monix.execution.Scheduler
 import outwatch.Sink.{ObservableSink, SubjectSink}
 import outwatch.dom.Observable
 
@@ -17,7 +18,7 @@ object Pipe {
     * @tparam T the type parameter of the elements
     * @return the newly created Pipe.
     */
-  def create[T](seeds: T*): IO[Pipe[T, T]] = create[T].map { pipe =>
+  def create[T](seeds: T*)(implicit s: Scheduler): IO[Pipe[T, T]] = create[T].map { pipe =>
     if (seeds.nonEmpty) {
       pipe.transformSource(_.startWith(seeds))
     }
@@ -26,7 +27,7 @@ object Pipe {
     }
   }
 
-  def create[T]: IO[Pipe[T, T]] = IO {
+  def create[T](implicit s: Scheduler): IO[Pipe[T, T]] = IO {
     val subjectSink = SubjectSink[T]()
     Pipe(subjectSink, subjectSink)
   }
--- a/src/main/scala/outwatch/Sink.scala
+++ b/src/main/scala/outwatch/Sink.scala
@@ -1,9 +1,9 @@
 package outwatch
 
 import cats.effect.IO
-import monix.execution.Scheduler.Implicits.global
+import monix.execution.Scheduler
 import monix.execution.{Ack, Cancelable}
-import monix.reactive.observers.Subscriber
+import monix.reactive.observers.{SafeSubscriber, Subscriber}
 import monix.reactive.subjects.PublishSubject
 import monix.reactive.{Observable, Observer}
 
@@ -22,7 +22,7 @@ sealed trait Sink[-T] extends Any {
     observable.subscribe(observer)
   }
 
-  private[outwatch] def observer: Observer[T]
+  private[outwatch] def observer: Subscriber[T]
 
   /**
     * Creates a new sink. That sink will transform the values it receives and then forward them along to this sink.
@@ -56,10 +56,10 @@ object Sink {
     override def unsafeSubscribeFn(subscriber: Subscriber[O]): Cancelable = stream.unsafeSubscribeFn(subscriber)
   }
 
-  private[outwatch] final case class SubjectSink[T]() extends Observable[T] with Sink[T] {
+  private[outwatch] final case class SubjectSink[T]()(implicit scheduler: Scheduler) extends Observable[T] with Sink[T] {
     private val subject = PublishSubject[T]
 
-    override private[outwatch] def observer = subject
+    override private[outwatch] def observer = SafeSubscriber(Subscriber(subject, scheduler))
 
     override def unsafeSubscribeFn(subscriber: Subscriber[T]): Cancelable = subject.unsafeSubscribeFn(subscriber)
   }
@@ -77,7 +77,7 @@ object Sink {
   def create[T](next: T => IO[Future[Ack]],
                 error: Throwable => IO[Unit] = _ => IO.pure(()),
                 complete: () => IO[Unit] = () => IO.pure(())
-               ): Sink[T] = {
+               )(implicit s: Scheduler): Sink[T] = {
     val sink = ObserverSink(
       new Observer[T] {
         override def onNext(t: T): Future[Ack] = next(t).unsafeRunSync()
@@ -111,6 +111,7 @@ object Sink {
     * @return the resulting sink, that will forward the values
     */
   def redirect[T,R](sink: Sink[T])(project: Observable[R] => Observable[T]): Sink[R] = {
+    implicit val scheduler = sink.observer.scheduler
     val forward = SubjectSink[R]()
 
     completionObservable(sink)
@@ -133,6 +134,7 @@ object Sink {
     * @return the two resulting sinks, that will forward the values
     */
   def redirect2[T,U,R](sink: Sink[T])(project: (Observable[R], Observable[U]) => Observable[T]): (Sink[R], Sink[U]) = {
+    implicit val scheduler = sink.observer.scheduler
     val r = SubjectSink[R]()
     val u = SubjectSink[U]()
 
@@ -159,6 +161,7 @@ object Sink {
   def redirect3[T,U,V,R](sink: Sink[T])
                        (project: (Observable[R], Observable[U], Observable[V]) => Observable[T])
                        :(Sink[R], Sink[U], Sink[V]) = {
+    implicit val scheduler = sink.observer.scheduler
     val r = SubjectSink[R]()
     val u = SubjectSink[U]()
     val v = SubjectSink[V]()
@@ -186,4 +189,6 @@ object Sink {
 
 }
 
-final case class ObserverSink[-T](observer: Observer[T]) extends AnyVal with Sink[T]
+final case class ObserverSink[-T](obs: Observer[T])(implicit s: Scheduler) extends Sink[T] {
+  override val observer = Subscriber(obs, s)
+}
--- a/src/main/scala/outwatch/Sink.scala
+++ b/src/main/scala/outwatch/Sink.scala
@@ -1,12 +1,14 @@
 package outwatch
 
 import cats.effect.IO
-import rxscalajs.facade.SubjectFacade
-import rxscalajs.subscription.Subscription
-import rxscalajs.{Observable, Observer, Subject}
+import monix.execution.Scheduler.Implicits.global
+import monix.execution.{Ack, Cancelable}
+import monix.reactive.observers.Subscriber
+import monix.reactive.subjects.PublishSubject
+import monix.reactive.{Observable, Observer}
 
+import scala.concurrent.Future
 
-import scala.scalajs.js
 
 sealed trait Sink[-T] extends Any {
 
@@ -16,7 +18,7 @@ sealed trait Sink[-T] extends Any {
     * Using this method is inherently impure and can cause memory leaks, if subscription
     * isn't handled correctly. For more guaranteed safety, use Sink.redirect() instead.
     */
-  def <--(observable: Observable[T]): IO[Subscription] = IO {
+  def <--(observable: Observable[T]): IO[Cancelable] = IO {
     observable.subscribe(observer)
   }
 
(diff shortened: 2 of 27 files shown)


src/test/scala/outwatch/ScenarioTestSpec.scala
```diff
@@ -193,38 +200,61 @@ class ScenarioTestSpec extends JSDomSpec {
...
+      todo <- Task {
+        val todo = "fold laundry"
+        inputElement.value = todo
+        inputElement.dispatchEvent(inputEvt)
+        submitButton.dispatchEvent(clickEvt)
+        todo
+      }
+      _ <- Task {
+        list.childElementCount shouldBe 1
+      }.executeWithFork
```
mariusmuja: This is because Monix uses by default a [batched execution model](https://monix.io/docs/2x/execution/scheduler.html#execution-model), so we need to "fork" the assertion to the end of the javascript execution queue, otherwise it might execute before all the effects of the event dispatch execute.

fdietze:
Would it be worth it to port directly to a preliminary [Monix 3](https://github.com/monix/monix/releases/tag/v3.0.0-M2) Release?

mariusmuja:
All that's needed to upgrade to Monix 3 is to update the dependency version, there seem to be no breaking changes in the API used.

fdietze:
Interesting. Are `monix.Task` and `cats.effect.IO` compatible?

LukaJCB:
They are!

src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -143,27 +142,29 @@ private[outwatch] final case class Receivers(
...
         .groupBy(_.attribute)
         .values
-        .map(_.last.attributeStream)(breakOut)
+        .map(_.last.attributeStream)(breakOut): _*
     )
 
-    attributeReceivers.combineLatest(childStreamReceivers)
+    attributeReceivers.startWith(Seq(Seq.empty)).combineLatest(
+      childStreamReceivers.startWith(Seq(Seq.empty))
+    ).dropWhile { case (a, c) => a.isEmpty && c.isEmpty }
```
fdietze: What is this `dropWhile` good for?

src/main/scala/outwatch/util/Store.scala
```diff
@@ -1,33 +1,37 @@
...
 import outwatch.dom._
 import outwatch.dom.helpers.STRef
-import rxscalajs.Observable
-import rxscalajs.subscription.Subscription
+
+import scala.concurrent.Future
+
+
 
 final case class Store[State, Action](initialState: State,
```
fdietze: `STRef` is only used in Store. Is it possible to remove it entirely?

src/main/scala/outwatch/http/Http.scala
```diff
@@ -1,10 +1,40 @@
...
+import org.scalajs.dom.ext.{Ajax, AjaxException}
+import org.scalajs.dom.{Blob, XMLHttpRequest}
+
+import scala.concurrent.Future
+import scala.scalajs.js
+import scala.scalajs.js.typedarray.ArrayBuffer
+import scala.scalajs.js.|
+
 
 object Http {
```
fdietze: It seems like here is more going on than a simple port. Could you explain what you did here? I'm also not using the `Http` module myself, so this should be reviewed by someone who does.

If its too different it should be a separate commit.

src/test/scala/outwatch/DomEventSpec.scala
```diff
@@ -418,8 +419,11 @@ class DomEventSpec extends JSDomSpec {
 
     document.getElementById("input").dispatchEvent(inputEvt)
 
-    winClicked shouldBe true
-    docClicked shouldBe true
+    Task {
```
fdietze: Could you please explain, why this is neccessary here?

src/main/scala/outwatch/Sink.scala
```diff
@@ -1,12 +1,14 @@
 package outwatch
 
 import cats.effect.IO
-import rxscalajs.facade.SubjectFacade
-import rxscalajs.subscription.Subscription
-import rxscalajs.{Observable, Observer, Subject}
+import monix.execution.Scheduler.Implicits.global
```
fdietze: Is it safe to import the global execution scheduler everywhere?

src/test/scala/outwatch/ScenarioTestSpec.scala
```diff
@@ -37,12 +38,18 @@ class ScenarioTestSpec extends JSDomSpec {
 
     document.getElementById("minus").dispatchEvent(event)
 
-    document.getElementById("counter").innerHTML shouldBe (-1).toString
-
-    for (i <- 0 to 10) {
-      document.getElementById("plus").dispatchEvent(event)
-      document.getElementById("counter").innerHTML shouldBe i.toString
-    }
+    Task {
```
fdietze: This whole block looks complicated compared to the old one. Is there a simpler way to write it?

fdietze:
@mariusmuja do you know how the `fullOpt` build size is affected?

src/main/scala/outwatch/dom/helpers/DomUtils.scala
mariusmuja: Without doing `startwith(Seq(Seq.empty))` and `dropWhile`, the `combinedLatest` would emit an element only when *both* the `attributeReceivers` and `childStreamReceivers` emit. This makes it emit when *either* of them emits.

src/main/scala/outwatch/util/Store.scala
mariusmuja: `STRef` is used in cases where a global variable with a referential transparent API is wanted (such as the store). It's possible to remove it, but without it the store would need to be passed as an argument to all components that use it.

src/main/scala/outwatch/http/Http.scala
mariusmuja: `rxscala` has an `Observable.ajax` method. Monix doesn't have that, so essentially I had to re-implement that functionality (based on `org.scalajs.dom.ext.Ajax`).

src/test/scala/outwatch/DomEventSpec.scala
mariusmuja: This is because Monix uses by default a [batched execution model](https://monix.io/docs/2x/execution/scheduler.html#execution-model), so we need to "fork" the assertion to the end of the javascript execution queue, otherwise it might execute before all the effects of the event dispatch execute.

src/main/scala/outwatch/Sink.scala
mariusmuja: It is ok, but it makes the [batched execution model](https://monix.io/docs/2x/execution/scheduler.html#execution-model) the default everywhere. It would be better to have the scheduler an implicit argument where it's needed, but that would have been a PR with more changes vs the rxscala version (which essentially doesn't have an  option of picking a different Scheduler).

src/test/scala/outwatch/ScenarioTestSpec.scala
mariusmuja: The simpler way would be to have the scheduler injected and use one with a synchronous execution model. Then the test code would be identical as before.

mariusmuja:
> @mariusmuja do you know how the fullOpt build size is affected?

Because of the dead code elimination that Scala.js does, it's difficult to say precisely. But assuming that no other features of Monix are used in addition to what the library (and the tests) are using, the bundled size seems to be smaller. I did a comparison of the test bundle size a while ago and posted it in #96 , here it is:

With RxJS:
```
-rw-rw-r--   1 marius marius  313987 Nov 21 21:41 outwatch-test-opt-bundle.js.gz
-rw-rw-r--   1 marius marius 1593988 Nov 21 21:34 outwatch-test-opt-bundle.js
-rw-rw-r--   1 marius marius  837289 Nov 21 21:33 outwatch-test-opt.js
```
Monix port:
```
-rw-rw-r--   1 marius marius  240514 Nov 21 21:38 outwatch-test-opt-bundle.js.gz
-rw-rw-r--   1 marius marius 1077036 Nov 21 21:38 outwatch-test-opt-bundle.js
-rw-rw-r--   1 marius marius 1046342 Nov 21 21:37 outwatch-test-opt.js
```

src/test/scala/outwatch/ScenarioTestSpec.scala
mariusmuja: Actually, the simplest way would be to have the synchronous scheduler everywhere (then it would be the same execution model as rxjs and the test could have stayed the same). But I don't like that, the batched execution model is a better choice to keep the page responsive.

src/main/scala/outwatch/dom/helpers/DomUtils.scala
fdietze: So the behavior of combineLatest is different in monix?

mariusmuja: That's what I thought too, but from reading the documentation now, it seems that `combineLatest` 
 in rxjs should behave the same as in Monix. Let me look into why I had to do this (or why it's not needed with rxjs/rxscala).

src/test/scala/outwatch/ScenarioTestSpec.scala
fdietze: does the test fail with the batched execution model?

fdietze:
I'm also a bit concerned that the users now have to deal with both Monix Tasks and IO. You said that they are compatible? But could outwatch be used with one of them only? How about IO and the batched execution context?

src/test/scala/outwatch/ScenarioTestSpec.scala
mariusmuja: Actually, it was not the batched execution model that was causing the async boundary here, it was the `Observable.merge` that uses a buffered subscriber to properly implement the back-pressure protocol. I've replaced it with a simple `mergeSync`, so the test stays unchanged.

mariusmuja:
> I'm also a bit concerned that the users now have to deal with both Monix Tasks and IO. You said that they are compatible? But could outwatch be used with one of them only? How about IO and the batched execution context?

Users don't have to deal with both. For example, `Task` is no longer used in the outwatch code (and before it was only used in the tests). 
IO doesn't know or care about the batched execution context, it's only Monix that does.

src/main/scala/outwatch/dom/helpers/DomUtils.scala
mariusmuja: The difference is between the  `Observable.combineLatest` rxscala companion object method and the Monix `Observable.combineLatestList`. The Monix `combineLatestList` doesn't emit an empty list if the argument list is empty, rxscala does.

mariusmuja:
The `Scheduler` is now injected as an implicit argument where needed. For example, this allows a `TrampolineScheduler` to be injected in the two tests where we want synchronous behaviour for easier testing.

Client code will require an implicit `Scheduler` in scope, but users will get a compile error asking to import `monix.execution.Scheduler.Implicits.global` (similar to the requirement to have an  `ExecutionContext` in scope when working with futures... in fact a `Scheduler` is an `ExecutionContext`).

src/main/scala/outwatch/dom/helpers/DomUtils.scala
fdietze: Should we have a test for this?

mariusmuja: Now the code doesn't do `.startWith(Seq(Seq.empty))....dropWhile { case (a, c) => a.isEmpty && c.isEmpty }` any more, instead it emits an `Observable(Seq.empty[Attribute])` if the `attributeReceivers` list is empty (just like in the rxscala version). There are several tests that fail if we don't do that.

src/main/scala/outwatch/Pipe.scala
```diff
@@ -16,16 +18,16 @@ object Pipe {
     * @tparam T the type parameter of the elements
     * @return the newly created Pipe.
     */
-  def create[T](seeds: T*): IO[Pipe[T, T]] = create[T].map { pipe =>
+  def create[T](seeds: T*)(implicit s: Scheduler): IO[Pipe[T, T]] = create[T].map { pipe =>
```
LukaJCB: I wonder, since we now have both `IO` and `Task` in our project as dependencies, if we should continue using one or the other, or abstract over them with the type classes found in `cats-effect`, i.e. 
```scala
def create[F[_]: Sync, T](seeds: T*): F[Pipe[T, T]]
```
What do you guys think? :)

On an unrelated note, I also think we could get rid of the varargs here, I don't think it really ever makes sense to prepend items to a stream?

fdietze: 1. I'm in favor of abstracting over IO and Task, this would also allow using outwatch without IO Monads again.

2. One seed item is definitely enough for state-management, which is actually what we need for reactive UIs ;) (#96)

@LukaJCB could you please open PRs/Issues for this? I'll merge now.

mariusmuja: +1 on abstracting over `IO` and `Task`.

LukaJCB: I created #135 and #136 :)

src/main/scala/outwatch/Sink.scala
```diff
@@ -48,35 +50,39 @@ sealed trait Sink[-T] extends Any {
...
-    override private[outwatch] def observer = sink.observer
+  private[outwatch] final case class ObservableSink[-I, +O](oldSink: Sink[I], stream: Observable[O]) extends Observable[O] with Sink[I] {
+    override private[outwatch] def observer = oldSink.observer
+
+    override def unsafeSubscribeFn(subscriber: Subscriber[O]): Cancelable = stream.unsafeSubscribeFn(subscriber)
   }
 
-  private[outwatch] final case class SubjectSink[T]() extends Subject[T](new SubjectFacade) with Sink[T] {
-    override private[outwatch] def observer = this
+  private[outwatch] final case class SubjectSink[T]()(implicit scheduler: Scheduler) extends Observable[T] with Sink[T] {
```
LukaJCB: Why exactly do we need the scheduler here? Could we not add it to the methods? This would unclutter our code a bit, no?

LukaJCB: In fact, why don't we do:

`def observer: Observer[T]`? So we wouldn't need the scheduler at all?

mariusmuja: We need it to create the `Subscriber` inside the `SubjectSink` (the `Sink.observer` is actually a `Subscriber` now) . If I would have left it an `Observer`, all the `Sink.redirect` operations would have required a `Scheduler` (and also many operations in `PipeOps`). This way they just use the `Scheduler` from the source `Sink`'s `Subscriber`.

LukaJCB: Yeah, that makes sense, however, I think creating a new `Handler` or `Pipe` is going to be a much much more common operation, then `redirect`, so I think it'd be better if the individual methods required the implicit evidence, no?

mariusmuja: Not quite, `EmitterBuilder`'s `-->` would require an implicit `Scheduler` in that case.

Issue #124, opened 2017-12-11
Global namespace


mariusmuja:
This has been discussed before in #32, but I'd like to bring up the issue again.

Currently, with the inclusion of `scala-dom-types` there are even more identifiers (html tags, attributes, events + css style names) imported into the global namespace by `import outwatch.dom._`. I think this is likely to cause issues along the way.

How do people feel about breaking compatibility and moving all dom identifiers to `outwatch.dom.all._`, so the name explicitly shows that *everything* gets imported and have `import outwatch.dom._` only import `Attributes`, `Tags`, `Styles`, `Handler`...

raquo:
So... what would be in `dom.all` but not in `dom`?

mariusmuja:
> So... what would be in dom.all but not in dom?

 `import outwatch.dom.all_` would bring in all html tags, attributes, styles, events, essentially what `import outwatch.dom._` does now and `import outwatch.dom._` would only import the `Attributes`, `Tags`, `Styles`... objects, so one could do
```scala
import outwatch.dom.{Attributes => A, Tags => T, Styles => S}

T.div( A.id := "some-id")
```

I realise that it's currently possible to do `import outwatch.dom.{Attributes => A, Tags => T, Styles => S}` and not use `import outwatch.dom._`, but I'm guessing that it's not what typically users are doing and having the `all` in the import makes it more obvious that *all* identifiers get imported, so it should be "handled with care" (maybe used only locally where the VDom is constructed).

cornerman:
> I realise that it's currently possible to do import outwatch.dom.{Attributes => A, Tags => T, Styles => S} and not use import outwatch.dom._,

Yes, but then you will need the implicits for working with `VDomModifier` and `VNode`. Therefore, I agree with this proposal. We should have an import `outwatch.dom._`, which we can advise people to use without filling up the scope so aggressively. They can still explicitly import all or build up a more fine granular scheme.

mariusmuja:
> Yes, but then you will need the implicits for working with VDomModifier and VNode.

Yes, good point!

mariusmuja:
If we decide to break compatibility like this, this would be an easy migration command for most situations:
```bash
find src -name '*.scala' | xargs sed -i -e 's/^\([ \t]*\)import outwatch.dom._/\1import outwatch.dom._\n\1import outwatch.dom.all._/'
```

mariusmuja:
Fixed by #132.

PR #132, opened 2017-12-14
DOM dsl refactor

mariusmuja:
I did some refactoring to make the DOM DSL imports more consistent and work better with IDE auto-completion.

To import all the DOM DSL identifiers (except for `TagsExtra` and `StylesExtra`) the following import needs to be used:
```
import outwatch.dom.dsl._
```

It's also possible to import just parts of the DSL, and IDE auto-completion helps with this:
```scala
import outwatch.dom.dsl.tags._
import outwatch.dom.dsl.attributes._
import outwatch.dom.dsl.styles._
```

This also addresses #124 by removing `Attributes` and `Tags` from the `outwatch.dom` namespace. Most existing code bases that use `import outwatch.dom._` can be upgraded with:
```bash
find src -name '*.scala' | xargs sed -i -e 's/^\([ \t]*\)import outwatch.dom._/\1import outwatch.dom._\n\1import outwatch.dom.dsl._/'
```

This PR is on top of #130.

### DIFF
--- a/src/main/scala/outwatch/dom/DomTypes.scala
+++ b/src/main/scala/outwatch/dom/DomTypes.scala
@@ -46,15 +46,6 @@ private[outwatch] object DomTypesBuilder {
       Cancelable(() => target.removeEventListener(key, eventHandler))
     }
   }
-
-  trait SimpleStyleBuilder extends StyleBuilders[IO[Style]] {
-
-    implicit def StyleIsBuilder[T](style: keys.Style[T]): StyleBuilder[T] = new StyleBuilder[T](style.cssName)
-
-    override protected def buildDoubleStyleSetter(style: keys.Style[Double], value: Double): IO[Style] = style := value
-    override protected def buildIntStyleSetter(style: keys.Style[Int],value: Int): IO[Style] = style := value
-    override protected def buildStringStyleSetter(style: keys.Style[_],value: String): IO[Style] = new StyleBuilder[Any](style.cssName) := value
-  }
 }
 import DomTypesBuilder._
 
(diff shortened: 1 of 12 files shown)


src/main/scala/outwatch/dom/DomTypes.scala
```diff
@@ -65,26 +56,34 @@ trait Tags
...
 
 trait Attributes
   extends Attrs
   with ReflectedAttrs
   with Props
   with Events
-  with Styles
-  with AttributesCompat
+  with AttributesExtra
+  with AttributeHelpers
```
cornerman: minor nitpick: can we merge `AttributesExtra` and `AttributeHelpers`

mariusmuja: Yes, good idea!

cornerman:
Nice, just looked at your last commit.

Can we keep the possibility of having just one import (like `import outwatch.dom.all._`)? Though, we should also be careful with introducing too many instances inheriting the same dom type traits, as this will lead to having duplicate inherited `lazy val`s in the memory (if both are used and not optimized away). Not sure...

mariusmuja:
The `import outwatch.dom.dsl._` imports everything (essentially what `import outwatch.dom.all._` was supposed to be).

cornerman:
But you additionally need `import outwatch.dom._`, right?

mariusmuja:
> But you additionally need import outwatch.dom._, right?

Oh, I see, that's what you meant. Yes, we still need that.

mariusmuja:
Everything can be imported as: `import outwatch.dom._, dsl._`

src/main/scala/outwatch/dom/dsl.scala
```diff
@@ -0,0 +1,24 @@
+package outwatch.dom
+
+object dsl extends Attributes with Tags with Styles {
+  object tags extends Tags {
+    object extra extends TagsExtra
+    object all extends Tags with TagsExtra
```
cornerman: I think `import dsl.tags._, extra._` is good enough for the rare(?) use case of needing extra tags.

src/main/scala/outwatch/dom/dsl.scala
```diff
@@ -0,0 +1,24 @@
...
+    object attrs extends Attrs
+    object reflected extends ReflectedAttrs
+    object props extends Props
+    object events extends Events
+    object outwatch extends OutwatchAttributes
+    object lifecycle extends OutWatchLifeCycleAttributes
+  }
+  object styles extends Styles {
+    object extra extends StylesExtra
+    object all extends Styles with StylesExtra
```
cornerman: same as for `tags.all`

cornerman:
> Everything can be imported as: import outwatch.dom._, dsl._

Good enough, because less lazy vals.

PR #133, opened 2017-12-14
Extended styles

mariusmuja:
This PR adds support for the Snabbdom [`delayed`, `remove` and `destroy` styles](https://github.com/snabbdom/snabbdom#delayed-properties) which allow for easy CSS animations. It also adds support for accumulating styles in order to be able to write composable animations like so:
```scala
 val slideInOut = VDomModifier(
    transition.accum := "transform .2s ease-in-out",
    transform := "translateX(-50px)",
    transform.delayed := "translateX(0px)",
    transform.remove := "translateX(50px)",
  )

  val fadeInOut = VDomModifier(
    transition.accum :=  "opacity .2s ease-in-out",
    opacity := 0,
    opacity.delayed := 1,
    opacity.remove := 0
  )

  val moveInOut =  VDomModifier(slideInOut, fadeInOut)
```

You can see this animation in action [here](https://mariusmuja.github.io/outwatch-demo-spa/#todo) when adding or removing items from the todo list.

This PR is on top of #132.

### DIFF
--- a/src/main/scala/outwatch/dom/DomTypes.scala
+++ b/src/main/scala/outwatch/dom/DomTypes.scala
@@ -136,10 +136,9 @@ abstract class DocumentEvents
 
 
 private[outwatch] trait SimpleStyleBuilder extends StyleBuilders[IO[Style]] {
-
   override protected def buildDoubleStyleSetter(style: keys.Style[Double], value: Double): IO[Style] = style := value
   override protected def buildIntStyleSetter(style: keys.Style[Int],value: Int): IO[Style] = style := value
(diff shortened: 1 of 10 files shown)


fdietze:
I really like the animation support!
Does the snabbdom delay also work with classes?

mariusmuja:
> Does the snabbdom delay also work with classes? 

No, it seems to be a feature of the style module:
https://github.com/snabbdom/snabbdom/blob/master/src/modules/style.ts

mariusmuja:
Rebased on master.

PR #139, opened 2017-12-26
CurrentTarget / Target ops

mariusmuja:
This PR is a result of the discussion we had in https://github.com/OutWatch/outwatch/pull/119#issuecomment-351861345. It still uses the non-typesafe cast for `currentTarget`, but allows for a nicer syntax and uses the `TagWith...` typeclasses for restricting the type the target can be cast to.

The syntax is the following:
 * `currentTarget` uses `html.Input` by default as the underlying event target:
```scala
input(
  onInput.value --> string,
  onClick.checked --> bool,
  onKeyDown.valueAsNumber --> number
)
```
The PR removes `onInputString`, `onInputNumber` and `onInputChecked` and updates the deprecation message on `inputString`, `inputNumber` and `inputChecked` to use  `onInput.value`, `onInput.valueAsNumber` and `onChange.checked` instead.

 *  it's possible to force a different event target for `currentTarget`, but only if the `TagWith...` typeclasses are implemented for it (`html.TextArea`, `html.Select` ).
 ```scala
textArea(
  onChange.value[html.TextArea] --> string,
)
```
* for `target`, the type from `TypedTargetEvent` is used by default
```scala
input(
  onSearch.target.value --> stringStream,
  onSearch.target.valueAsNumber --> doubleStream,
  onSearch.target.checked --> boolStream,
)
```
* in case the type of `TypedTargetEvent` is not specific enough it's possible to force a more specific type:
```scala
input(
  //onClick.target.value --> stringStream,   // does not compile because onClick is TypedTargetEvent[dom.Element], not specific enough
  onClick.target[html.Input].value --> stringStream,
  onClick.target[html.Input].valueAsNumber --> doubleStream,
  onChange.target[html.Input].checked --> boolStream,
)
```

### DIFF
--- a/src/main/scala/outwatch/dom/helpers/TargetOps.scala
+++ b/src/main/scala/outwatch/dom/helpers/TargetOps.scala
@@ -1,0 +1,24 @@
+package outwatch.dom.helpers
+
+import org.scalajs.dom.{Event, html}
+
+trait TargetOps {
+
+  implicit class TargetAsInput[E <: Event, O <: Event](builder: EmitterBuilder[E, O]) {
+
+    object target {
+      def value: EmitterBuilder[E, String] = builder.map(_.target.asInstanceOf[html.Input].value)
+
+      def valueAsNumber: EmitterBuilder[E, Int] = builder.map(_.target.asInstanceOf[html.Input].valueAsNumber)
+
+      def checked: EmitterBuilder[E, Boolean] = builder.map(_.target.asInstanceOf[html.Input].checked)
+    }
+
+    def value: EmitterBuilder[E, String] = builder.map(e => e.currentTarget.asInstanceOf[html.Input].value)
+
+    def valueAsNumber: EmitterBuilder[E, Int] = builder.map(e => e.currentTarget.asInstanceOf[html.Input].valueAsNumber)
+
+    def checked: EmitterBuilder[E, Boolean] = builder.map(e => e.currentTarget.asInstanceOf[html.Input].checked)
+  }
+
+}
--- a/src/test/scala/outwatch/DomEventSpec.scala
+++ b/src/test/scala/outwatch/DomEventSpec.scala
@@ -286,9 +286,10 @@ class DomEventSpec extends JSDomSpec {
   it should "be able to be transformed from strings" in {
 
     val number = 42
+    val onInputValue = onInput.value
     val node = Handler.create[Int].flatMap { stream =>
       div(
-        button(id := "input", onInputString(number) --> stream),
+        button(id := "input", onInputValue(number) --> stream),
         span(id:="num",child <-- stream)
       )
     }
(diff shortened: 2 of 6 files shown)


cornerman:
Thanks for working on this. I did not have time to make a full review of the changes just yet.

But I do not get why we need these generic methods for currentTarget `onChange.value[html.TextArea]`. Just using `onChange.value` will behave the same way, as casting is unchecked in scalajs anyhow. What is the benefit of having it?

Also, why do we decide that `currentTarget` can assume `html.Input`, but `target` cannot? Here, we do not have typesafety anyhow.

mariusmuja:
> But I do not get why we need these generic methods for currentTarget `onChange.value[html.TextArea]`. Just using `onChange.value` will behave the same way, as casting is unchecked in scalajs anyhow. What is the benefit of having it?
>
> Also, why do we decide that `currentTarget` can assume `html.Input`, but `target` cannot? Here, we do not have typesafety anyhow.

You are correct, there's no benefit from having the generic methods for casting to different event targets, they will behave the same way. I've removed them and simplified the PR.

src/test/scala/outwatch/DomEventSpec.scala
```diff
@@ -426,4 +427,69 @@ class DomEventSpec extends JSDomSpec {
     docClicked shouldBe true
   }
 
+  "TagWith" should "correctly work on events" in {
```
cornerman: rename TagWith to TargetOps in test string here

src/main/scala/outwatch/dom/helpers/TargetOps.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  implicit class CurrentTargetAsInput[E <: Event, O <: Event](builder: EmitterBuilder[E, O]) {
+    def value: EmitterBuilder[E, String] = builder.map(e => e.currentTarget.asInstanceOf[html.Input].value)
+
+    def valueAsNumber: EmitterBuilder[E, Int] = builder.map(e => e.currentTarget.asInstanceOf[html.Input].valueAsNumber)
+
+    def checked: EmitterBuilder[E, Boolean] = builder.map(e => e.currentTarget.asInstanceOf[html.Input].checked)
+  }
+
+  implicit class TargetMap[E <: Event, O <: Event](builder: EmitterBuilder[E, O]) {
```
cornerman: I would argue against mapping to EventTarget here, as `_.map(_.target)` seems good enough to me and we do not have the same helper functions for other event properties. Instead we can only define a target object with the above methods in an implicit class for the EmitterBuilder.

Like this:
```scala
  implicit class TargetAsInput[E <: Event, O <: Event](builder: EmitterBuilder[E, O]) {
      object target {
          def value = ???
          def valueAsNumber = ???
          def checked = ???
      }
￼   }
```

mariusmuja: Good point, I'll update the PR.

Issue #137, opened 2017-12-24
Partially rendering components


fdietze:
Annoyances like in #134 only occur, because we abuse an event library to represent (UI) state. We need workarounds because we allow an inconsistent state to be defined in the first place (instead of requiring a default value in each variable). This results in inserting fake-content in the form of empty string-nodes or absent attributes which is not always the desired solution. If it is not totally clear what I'm saying here, please ask.

It is possible to define further, similarly failing tests.

```scala
  it should "partially render component even if parts not present" in {
    val messagesColor = PublishSubject[String]
    val messagesBgColor = PublishSubject[String]
    val childString = PublishSubject[String]

    val vNode = div(
      color <-- messagesColor,
      backgroundColor <-- messagesBgColor,
      child <-- childString
    )

    val node = document.createElement("div").asInstanceOf[html.Div]
    document.body.appendChild(node)
    OutWatch.renderInto(node, vNode).unsafeRunSync()

    node.innerHTML shouldBe "<div></div>"
    node.style.color shouldBe ""
    node.style.backgroundColor shouldBe ""

    childString.onNext("fish")
    node.innerHTML shouldBe "<div>fish</div>"
    node.style.color shouldBe ""
    node.style.backgroundColor shouldBe ""

    messagesColor.onNext("red")
    node.innerHTML shouldBe "<div>fish</div>"
    node.style.color shouldBe "red"
    node.style.backgroundColor shouldBe ""

    messagesBgColor.onNext("blue")
    node.innerHTML shouldBe "<div>fish</div>"
    node.style.color shouldBe "red"
    node.style.backgroundColor shouldBe "blue"
  }
```

We have to ask ourselves if this should be the desired behavior. What to do instead, or if we can forbid inconsistent state via the type system. For Monix, see https://github.com/monix/monix/issues/279.

always related: #96

PR #140, opened 2017-12-28
Combine observables using merge/scan

mariusmuja:
This PR refactors the way the VNode observables are combined, using `merge` & `scan` (and explicit VNode state) instead of `combineLatest`.

I think it makes the code simpler and removes the need for workarounds such as keeping track if there's one or more child observables or making sure there's a `startWith` on the streams for partial rendering, avoiding issues such as #137, #134.

### DIFF
--- a/src/main/scala/outwatch/dom/helpers/DomUtils.scala
+++ b/src/main/scala/outwatch/dom/helpers/DomUtils.scala
@@ -1,7 +1,6 @@
 package outwatch.dom.helpers
 
 import cats.effect.IO
-import monix.reactive.subjects.BehaviorSubject
 import outwatch.dom._
 
 import scala.collection.breakOut
@@ -47,7 +46,7 @@ object Children {
 
     override def ::(node: ChildVNode): Children = node match {
       case s: StringVNode => toModifier(s) :: this
-      case n => n :: VNodes(Nil, StreamStatus())
+      case n => n :: VNodes(Nil, hasStream = false)
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -144,48 +136,48 @@ private[outwatch] final case class SeparatedEmitters(
...
-        case (csr: ChildrenStreamReceiver, obs) =>
-          obs.combineLatestMap(
-            if (childStreamStatus.hasMultipleChildOrChildren) csr.childrenStream.startWith(Seq(Seq.empty)) else csr.childrenStream
-          )((nodes, n) => n.toList ++ nodes)
-      }
-    } else {
-      Observable(Seq.empty)
-    }
-
-    // only use last encountered observable per attribute
```
fdietze: Do we have a test for this?

mariusmuja: This is very much at the core of the library, all tests that contain VDOM Observables are tests for it.

fdietze: I was referring to:
```
// only use last encountered observable per attribute
```
I was just wondering if we have a test for this, else I would write one.

mariusmuja: Yes, it's tested in a couple of places:

https://github.com/OutWatch/outwatch/blob/master/src/test/scala/outwatch/OutWatchDomSpec.scala#L744

https://github.com/OutWatch/outwatch/blob/master/src/test/scala/outwatch/OutWatchDomSpec.scala#L725

mariusmuja:
I'd like to get a second review on this before merging it.

fdietze:
@cornerman @LukaJCB

src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -56,33 +55,26 @@ object Children {
 
     override def ::(node: ChildVNode): Children = node match {
       case s: StringVNode => toModifier(s) :: this // this should never happen
-      case n => n :: VNodes(modifiers.map(toVNode), StreamStatus())
+      case n => n :: VNodes(modifiers.map(toVNode))
     }
   }
 
-  private[outwatch] case class VNodes(nodes: List[ChildVNode], streamStatus: StreamStatus) extends Children {
+  private[outwatch] case class VNodes(nodes: List[ChildVNode], hasStream: Boolean = false) extends Children {
```
cornerman: I would remove the default argument here, because you can easily create an invalid state here, so you should think about this argument when constructung this VNodes class.

cornerman:
@mariusmuja Could you kindly rebase on master? :)

mariusmuja:
Rebased on master.

Issue #77, opened 2017-11-05
Refactor LocalStorage util


fdietze:
1. Use the [Storage API](https://developer.mozilla.org/de/docs/Web/API/Storage) ([in scalajs.dom.ext](https://www.scala-js.org/api/scalajs-dom/0.9.0/#org.scalajs.dom.ext.Storage)) instead of LocalStorage.
2. Read and write `Option[String]` for non-existing keys and the ability for removing keys by writing `None`.

cornerman:
We should also complete the Observable [here](https://github.com/OutWatch/outwatch/blob/master/src/main/scala/outwatch/util/LocalStorage.scala#L11), after emitting the current value from storage.

PR #145, opened 2018-01-06
Make Handler and Pipe factory public

### LINKED ISSUE #143: Create Handler from Observable and Sink
Hey there,

I want to create a `Handler` that is manually composed from an `Observable[T]` and a `Sink[T]` of the same type. It seems like `ObservableSink` is what I want, but since it is private I can not access it.

In my example I want to create a Handler within a case class like this:
```
val hand: Handler[T] = {
  val obs: Observable[T] = myReader(Object.arg)
  val sink: Sink[T] = myWriter(Object.arg)
  Sink.ObservableSink(sink, obs)
}
```

So how can I achieve that? Why is `ObservableSink` or the `Handler` and `Pipe` factories private? 

Thanks for your help in advance!

Kind regards

GRBurst:
By making the `Pipe` and `Handler` factories public, it is possible to self-compose `Handler` or `Pipe` providing a `Sink` and an `Observable`. This offers much more flexibility.

Tests run without an issue.

@doofin Does that solve your issue, too?

fixes #143

### DIFF
--- a/src/main/scala/outwatch/Handler.scala
+++ b/src/main/scala/outwatch/Handler.scala
@@ -5,7 +5,7 @@ import monix.execution.Scheduler
 import outwatch.dom.Observable
 
 object Handler {
-  private[outwatch] def apply[T](sink: Sink[T], source: Observable[T]): Handler[T] = Pipe(sink, source)
+  def apply[T](sink: Sink[T], source: Observable[T]): Handler[T] = Pipe(sink, source)
 
   /**
     * This function also allows you to create initial values for your newly created Handler.
--- a/src/main/scala/outwatch/Pipe.scala
+++ b/src/main/scala/outwatch/Pipe.scala
@@ -7,7 +7,7 @@ import outwatch.dom.Observable
 
 
 object Pipe {
-  private[outwatch] def apply[I, O](sink: Sink[I], source: Observable[O]): Pipe[I, O] =
+  def apply[I, O](sink: Sink[I], source: Observable[O]): Pipe[I, O] =
     new ObservableSink[I, O](sink, source)
 
   /**


doofin:
thanks, would like to try that later!

mariusmuja:
In my mind `Pipe` has/should have a contract (currently unwritten):

1. if the `Sink` is completed, the `Source` will be completed.
1. if the `Sink` receives an error, the `Source` with emit the error
1. if something is sent into the `Sink` it's (probbaly) emitted by the `Source` (could be filtered or transformed)

For example the `Sink.redirect` ensures that the first item is of this contract is obeyed when doing a redirect on a Sink:
https://github.com/OutWatch/outwatch/blob/master/src/main/scala/outwatch/Sink.scala#L117

Mixing random `Observables` and `Sink` can break this contract. 

What is the use case when an `Observable` and a `Sink` should be mixed into a `Pipe` that cannot be accomplished by the operations in `PipeOps`: https://github.com/OutWatch/outwatch/blob/master/src/main/scala/outwatch/package.scala#L7
?

cornerman:
> In my mind Pipe has/should have a contract (currently unwritten):
> 
> 1. if the Sink is completed, the Source will be completed.
> 2. if the Sink receives an error, the Source with emit the error
> 3. if something is sent into the Sink it's (probbaly) emitted by the Source (could be filtered or transformed)

I agree with point 1 and 2, because especially when working with a lot of composed observables, this would be tedious to debug. For point 3, this is not really part of the contract as `transform` already allows this, as you also mention.

Could we maybe ensure completion and error behaviour when constructing an `ObservableSink`? Also, we would need to see whether the concrete use case can be done with `transform` arleady?

fdietze:
A concrete use-case would be implementing a Handler for the local storage:
https://github.com/OutWatch/outwatch/blob/master/src/main/scala/outwatch/util/LocalStorage.scala

How can this be done without the Handler-factory?

mariusmuja:
I don't think local storage should be a `Handler`.

fdietze:
Why not? It would be a convenient way to write into and subscribe to changes from local storage. The interface is like a regular handler, but the values are persisted.

mariusmuja:
Because the semantics are not of a handler (at least not they way it's implemented at the moment). The `Observable` emits a single element upon creation and nothing else even if other values are sent into the Sink. Also it wouldn't satisfy the contract I mentioned above. 

Thinking about it more, LocalStorage is pretty broken the way it's currently implemented. It could be turned into a `Handler` using the storage event (https://developer.mozilla.org/en-US/docs/Web/Events/storage), I'll attempt that in another PR.

mariusmuja:
> Very good, but can we keep both options? I do not think that you always want to subscribe to the event but just want to feed the current value once.

It would be as easy as `LocalStorage.take(1)` (or something like that).

fdietze:
Agreed, the way it's implemented now does not have the semantics of a Handler, but it can be done with storage events. I mentioned that before in #77.
My Point is, that one should be able to implement these kinds of Handlers in client code.

mariusmuja:
> My Point is, that one should be able to implement these kinds of Handlers in client code.

That's possible with the current API. Actually, an arbitrary handler/pipe can be constructed with the current API like this:
```
val obs: Observable[B] = ...
val pipe: IO[Pipe[A,B]] = Handler.create[A].map(_.transformSource(_ => obs))
```
When `transformSource` like this, a "handler" can be created that doesn't have handler semantics, but it's less trivial to do so compared to having a public `Pipe` constructor. Also, the `transformSource` name suggests that one should be transforming the existing observable rather than returning a new one.

fdietze:
How would I include a `Sink` in this example? `transformSink` only accepts `Observable[I2] => Observable[I]`.

mariusmuja:
Yes, you can't change the `Sink`, only redirect it. You have to create the `Sink` using `Pipe.create`, when you want to have a pipe (in order to have pipe semantics).

I think this discussion boils down to: is there a legitimate use case when the client code would need to create a `Pipe` (with the semantics of a Pipe) and it's not possible to do so using the operations in `PipeOps`.

fdietze:
After a discussion with @cornerman we came to the conclusion that the problem actually was how we can construct a `Handler` which interacts with a non-reactive interface, which for example provides a setter/writer-function and a callback. As it turns out, this can already be done with `foreach` and `unsafeOnNext`:
```scala
val h = Handler.create().unsafeRunSync()
h.foreach(v => setter(v)) //call setter function
// And let the callback call
h.unsafeOnNext(newValue)
```

fdietze:
For the record:
It should be possible to create such a factory like this:
```scala
def handlerFactory[T](o: Observable[T], s: Sink[T])(implicit scheduler: Scheduler): IO[Handler[T]] = {
  for {
    h <- Handler.create[T]()
  } yield {
    h.transformHandler[T](_ => o) { i => i.foreach(s.unsafeOnNext); Observable.empty }
  }
}
```

PR #146, opened 2018-01-06
Emitter accept side-effecting functions

fdietze:
One can now write things like:

```scala
button(
  onClick --> (event => handleEvent(event)),
  onClick(1) --> (counter += _),
  onClick --> (() => counter += 1),
)
```

### DIFF
--- a/src/main/scala/outwatch/dom/package.scala
+++ b/src/main/scala/outwatch/dom/package.scala
@@ -2,7 +2,7 @@ package outwatch
 
 import cats.effect.IO
 
-package object dom extends Implicits with ManagedSubscriptions {
+package object dom extends Implicits with ManagedSubscriptions with SideEffects {
 
   type VNode = IO[VTree]
   type VDomModifier = IO[VDomModifier_]
--- a/src/main/scala/outwatch/SideEffects.scala
+++ b/src/main/scala/outwatch/SideEffects.scala
@@ -1,0 +1,11 @@
+package outwatch
+
+import monix.execution.Scheduler
+import monix.execution.Ack.Continue
+import cats.effect.IO
+
+trait SideEffects {
+  def sideEffect[T](f: T => Unit)(implicit s: Scheduler): Sink[T] = Sink.create[T] { e => IO { f(e); Continue } }(s)
+  def sideEffect[S,T](f: (S,T) => Unit)(implicit s: Scheduler): Sink[(S,T)] = Sink.create[(S,T)] { e => IO { f(e._1, e._2); Continue } }(s)
+  def sideEffect(f: => Unit)(implicit s: Scheduler): Sink[Any] = Sink.create[Any] { e => IO { f; Continue } }(s)
+}
--- a/src/test/scala/outwatch/DomEventSpec.scala
+++ b/src/test/scala/outwatch/DomEventSpec.scala
@@ -304,6 +304,45 @@ class DomEventSpec extends JSDomSpec {
     document.getElementById("num").innerHTML shouldBe number.toString
   }
 
+  it should "handler can trigger side-effecting functions" in {
+    var triggeredEventFunction = 0
+    var triggeredIntFunction = 0
+    var triggeredFunction = 0
+    var triggeredFunction2 = 0
+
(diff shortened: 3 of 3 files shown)


mariusmuja:
I don't think we should have this in Outwatch, it encourages an imperative style of programming.

Plus, it's super easy to add this functionality to a client program if one really wants it (though an either implicit or explicit conversion from function to Sink).

fdietze:
I'm not sure if a library can *encourage* an imperative style of programming. In my opinion the motivation to write functional code doesn't come from the enforcement of a library, but from insights of the developer itself.

The code in my projects is 95% functional, but for some parts it's simply impractical to write functional code. Especially for websites. For these parts I want the libraries to help me to still be productive. Scala allows me to do exactly this. It's annoying if a library forces me to use workarounds for some ideological reasons.

What I'm proposing here is a shorter way to write this:

```scala
val handler = Handler.create[Int]()
var counter = 0
handler(counter += _)
button(
  onClick(1) --> handler,
)
```
It can now be written as:

```scala
var counter = 0
button(
  onClick(1) --> (counter += _),
)
```
Which is much easier to write and reason about.

In my project I had implicits for converting functions to sinks, but the type inference was a bit too bad. So I decided to put the support into the OutWatch.

mariusmuja:
> I'm not sure if a library can *encourage* an imperative style of programming. In my opinion the motivation to write functional code doesn't come from the enforcement of a library, but from insights of the developer itself.

I think a library can absolutely enourage a style of programming over another. It's true that the insights of the developer are most important, but a library can make it easier to use a style over another.

> It can now be written as:
> ```scala
> var counter = 0
> button(
>  onClick(1) --> (counter += _),
> )
> ```
> Which is much easier to write and reason about.

I find that harder to reason about. I expect something that's on the right of `-->` to be a sink.

> In my project I had implicits for converting functions to sinks, but the type inference was a bit too bad. 

A client code could use something like this for example if side effects are desired:

```scala
def sideEffect[T](f: T => Unit)(implicit s: Scheduler) =  Sink.create[T]{e => IO{f(e);  Continue}}(s)

var counter = 0
button(
  onClick(1) --> sideEffect(counter += _),
)
```
which is better then an implicit conversion in my opinion, because it states explicitly what is side effect.

fdietze:
Ok, you have valid arguments. Thanks for clarifying.

fdietze:
Here is a better use-case, why side-effects are important. We are still working with HTML:
```scala
onPostpatch --> sideEffect[(Element, Element)] { case (_, elem) => scrollToBottom(elem) }
```

cornerman:
I would like to see th `sideeffect` function in outwatch. What do you think?

It is often needed to issue sideeffects when working with the dom, where we need to react to changes/events. And creating a handler, registering the sideeffect in the handler and attaching the handler to an event seems is tedious to write in these cases.

mariusmuja:
> I would like to see the `sideEffect` function in outwatch. What do you think?

I'm ok with having the `sideEffect` function in outwatch.

fdietze:
Just pushed a commit adding this function.

src/main/scala/outwatch/dom/package.scala
```diff
@@ -23,4 +25,9 @@ package object dom extends Implicits with ManagedSubscriptions {
 
   type Handler[T] = outwatch.Handler[T]
   val Handler = outwatch.Handler
+
+
+  def sideEffect[T](f: T => Unit)(implicit s: Scheduler): Sink[T] = Sink.create[T] { e => IO { f(e); Continue } }(s)
+  def sideEffect[S,T](f: (S,T) => Unit)(implicit s: Scheduler): Sink[(S,T)] = Sink.create[(S,T)] { e => IO { f(e._1, e._2); Continue } }(s)
+  def sideEffect(f: => Unit)(implicit s: Scheduler): Sink[Any] = Sink.create[Any] { e => IO { f; Continue } }(s)
```
mariusmuja: It would be nicer if these were in a `SideEffects` trait rather than in the package object.

fdietze:
I moved the functions into a separate trait.

PR #147, opened 2018-01-08
Update dependencies: monix, cats, cats-effect, snabbdom

mariusmuja:
Updated some dependencies:

monix - 3.0.0-M3
cats - 1.0.1
cats-effect - 0.8
snabbdom - 0.7.1


LukaJCB:
👍

Issue #76, opened 2017-11-05
Provide imports for testing and large codebase migrations


fdietze:
Currently it is not possible to manually write data into a `Sink` or `Handler`. For safety reasons the `next()` method on the inner observable of Sink is hidden. This is a good thing to discourage the usage of side-effects.

Though this would be useful to use in tests.

Another use-case is migration of large code-bases. Allowing side-effecting parts of the code to write into `Sink`/`Handler` would help to at least compile a partially migrated code base.

I suggest an Import like
```scala
import outwatch.sinkUnsafeNext
```

where `SinkUnsafeNext` is an implicit class over Sink to provide an `unsafeNext()` method.

mariusmuja:
You could construct and ObserverSink that allows you to push items into it (has the inner observer public):
```scala
  def unsafeSink[T](sink: Sink[T]): IO[ObserverSink[T]] = {
    val subject = Subject[T]
    val newSink = ObserverSink(subject)

    (sink <-- subject).map(_ => newSink)
  }
```

For example:

```scala
def main(args: Array[String]): Unit = {

    val handler = Handlers.createStringHandler()

    handler.flatMap { handler =>
      handler(elem => dom.console.log(s"Elem: $elem"))

      unsafeSink(handler).map { sink =>
        sink.observer.next("One")
        sink.observer.next("Two")
        sink.observer.next("Three")
      }
    }.unsafeRunSync()
  }
```

cornerman:
But why is the `observer` in `ObserverSink` public, but not in `Sink`?

cornerman:
Fixed with #149

PR #149, opened 2018-01-10
Add unsafeOnNext to Sink

fdietze:
Another controversial feature! 😆 

I'm interfacing a lot with `d3`, so this and #146 are important to me. Implementing this function directly on `Sink` is way more efficient than implementing it using `unsafeSink` from #76.

### DIFF
--- a/src/main/scala/outwatch/Sink.scala
+++ b/src/main/scala/outwatch/Sink.scala
@@ -24,6 +24,8 @@ sealed trait Sink[-T] extends Any {
 
   private[outwatch] def observer: Subscriber[T]
 
+  def unsafeOnNext(value:T):Future[Ack] = observer.onNext(value)
+
   /**
     * Creates a new sink. That sink will transform the values it receives and then forward them along to this sink.
     * The transformation is described by a function from an Observable to another Observable, i.e. an operator on Observable.
--- a/src/test/scala/outwatch/HandlerSpec.scala
+++ b/src/test/scala/outwatch/HandlerSpec.scala
@@ -1,6 +1,14 @@
 package outwatch
 
 class HandlerSpec extends JSDomSpec {
+  "Sink" should "have unsafeOnNext" in {
+    val handler = Handler.create[Int].unsafeRunSync()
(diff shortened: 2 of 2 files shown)


LukaJCB:
I'm personally okay with this one, as most people won't want to use the "unsafe" functions.

src/main/scala/outwatch/Sink.scala
```diff
@@ -24,6 +24,8 @@ sealed trait Sink[-T] extends Any {
 
   private[outwatch] def observer: Subscriber[T]
 
+  def unsafeOnNext(value:T):Unit = { observer.onNext(value) }
```
mariusmuja: This should return `Future[Ack]`, so users can implement the back-pressure protocol.

mariusmuja: A more generic (and safer) version would be:
```scala
  def feed(values: T*): Future[Ack] = {
    Observer.feed(observer, values)
  }
```
This can feed multiple values into the observer while implementing back-pressure as well.

fdietze: Should I just take this function and rename it `unsafeOnNext`?

I'm a bit against varargs. What is the use-case for feeding multiple values in one run?

mariusmuja: No, I don't think this function should be named `unsafeOnNext`, I think `unsafeOnNext` should call `onNext` as in the PR, but have the correct return type (`Future[Ack]`).

I don't have a use case for it at the moment. I only suggested it because if somebody wanted to correctly feed multiple values with `unsafeOnNext`, he/she should handle the back-pressure protocol (only feed the next element when the future of the previous one completes with `Continue`) and `feed` does this.

fdietze: Ok. Let's tackle this once we have a use-case for the `feed` method. For now I'll just add `unsafeOnNext`.

fdietze:
`unsafeOnNext` now returns `Future[Ack]`.

PR #151, opened 2018-01-12
Hooks accept html.Element in Sinks

fdietze:
This allows to write side-effects on HTMLElement which provides all the expected methods:
```scala
input(
  onInsert --> sideEffect(_.focus())
)
```

Of course the hard casting is not optimal, but very convenient until we have a better option. It would be much better to accept Sinks of the real types of the corresponding HTML-Elements.

This is exactly the same problem as we are facing with `event.currentTarget` in #93 and #119. I think it is very important to achieve type-safety when interacting with the dom.


src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -188,4 +188,4 @@ private[outwatch] final case class Receivers(
   lazy val nonEmpty: Boolean = {
     attributeStreamReceivers.nonEmpty || childStreamStatus.hasChildOrChildren
   }
-}
\ No newline at end of file
+}
```
cornerman: remove noise from commit ;)

mariusmuja:
Outwatch/snabbdom could be used for SVG as well. This PR would break that.

cornerman:
> Outwatch/snabbdom could be used for SVG as well. This PR would break that.

What makes this different from the former `onInputString` and the new `event.(value|valueAsNumber|checked)` where we assume `InputEvent`. This is for the lack of knowledge about the embedding element type.

mariusmuja:
> with domtypes we are anyhow only creating html elements from tags, so the cast should be safe to assume.

SVG support could be added to scala-dom-types.

> What makes this different from the former onInputString and the new event.(value|valueAsNumber|checked) where we assume InputEvent. This is for the lack of knowledge about the embedding element type.

The fact that `onInputString` would not be used/useful for SVG elements, however hook events would be.

mariusmuja:
> But still, we already have that situation with, e.g., div(onInput.value --> something, contenteditable := true). The analogy was that something like this would equally thow a runtime exception.

However, the `onInput` event would not be triggered for SVG nodes, but hooks such as `onUpdate`, `onPostPatch` would.

cornerman:
> However, the onInput event would not be triggered for SVG nodes, but hooks such as onUpdate, onPostPatch would.

The example has nothing to do with SVG. Additionally, the api would not break for svg element, it would only break if html specific functions were called.

For the problem at hand, there needs to a be a convenient solution. Maybe just an extension method `asHTML` for the HookBuilder?

mariusmuja:
> For the problem at hand, there needs to a be a convenient solution. Maybe just an extension method asHTML for the HookBuilder?

Yes, an extension method would be better, maybe something like: `.as[html.Element]` ?

As an observation, when I needed functionality like this, I created a collection of side effects. Typically, I only need a couple of such side effects, so I can write them once and re-use them across the project. This would obviously not be practical in cases where there's a lot of such side effects:

```scala
object Hooks {

  val selectOnClick: VDomModifier = onInsert --> Sink.create[dom.Element] { e =>
    IO {
      e match {
        case input: html.Input => input.onclick = _ => input.select()
        case _ =>
      }
      Continue
    }
  }

  val focusAfterUpdate: VDomModifier = onPostpatch --> Sink.create[(dom.Element, dom.Element)] { case (_, e) =>
    IO {
      e match {
        case elem: html.Element => elem.focus()
        case _ =>
      }
      Continue
    }
  }
}
```

Then use them like:
```scala
input(Hooks.focusAfterUpdate, ...)
```

fdietze:
Ok. Now I think we should first tackle #152

fdietze:
closing in favor of #158

PR #156, opened 2018-01-16
Make Hooks and EmitterBuilder consistent

### LINKED ISSUE #152: Make hooks and emitterbuilder consistent
I just found myself writing `onInsert.map(...) --> ...`

Should we make the hooks also use the `EmitterBuilder` infrastructure? What do you think?

mariusmuja:
Makes `Hooks` and `EmitterBuilder` consistent, closes #152.

### DIFF
--- a/src/main/scala/outwatch/dom/DomTypes.scala
+++ b/src/main/scala/outwatch/dom/DomTypes.scala
@@ -17,6 +17,12 @@ import monix.reactive.OverflowStrategy.Unbounded
 
 import scala.scalajs.js
 
+private[outwatch] object Builders {
+  type Attribute[T, _] = helpers.ValueBuilder[T, Attr]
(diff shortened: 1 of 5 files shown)


src/main/scala/outwatch/dom/DomTypes.scala
```diff
@@ -17,6 +17,12 @@ import monix.reactive.OverflowStrategy.Unbounded
 
 import scala.scalajs.js
 
+private[outwatch] object Builders {
+  type Attribute[T, _] = helpers.ValueBuilder[T, Attr]
```
cornerman: maybe can also live in `DomTypesBuilders`?

fdietze: I'm fine with both. We'll see how it turns out in the next refactoring 😉

PR #157, opened 2018-01-16
Implement LocalStorage Handler which reacts on StorageEvent

### LINKED ISSUE #77: Refactor LocalStorage util

fdietze:
fixes #157 -- does github break if I do this? 😈 
fixes #77

### DIFF
--- a/src/main/scala/outwatch/util/LocalStorage.scala
+++ b/src/main/scala/outwatch/util/LocalStorage.scala
@@ -12,6 +12,28 @@ import cats.implicits._
 import outwatch.Handler
 
 object Storage {
+  private def handlerWithTransform(domStorage: dom.Storage, key: String, transform: Observable[Option[String]] => Observable[Option[String]])(implicit scheduler: Scheduler) = {
+    val storage = new dom.ext.Storage(domStorage)
+
+    for {
+      h <- Handler.create[Option[String]](storage(key))
+    } yield {
+      // We execute the write-action to the storage
+      // and pass the written value through to the underlying handler h
+      h.transformHandler(o => transform(o).distinctUntilChanged) { input =>
+        input.foreach {
+          case Some(data) => storage.update(key, data)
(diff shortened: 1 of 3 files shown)


src/main/scala/outwatch/util/LocalStorage.scala
```diff
@@ -1,30 +1,52 @@
...
-  def apply(key: String): Observable[String] = {
-    Observable.create[String](Unbounded){ observer =>
-      observer.onNext(localStorage.getItem(key))
-      Cancelable.empty
+    for {
+      h <- Handler.create[Option[String]](storage(key))
+    } yield {
+      // We execute the write-action to the storage
+      // and pass the written value through to the underlying handler h
+      h.transformHandler(Observable.merge(_, storageEvents).distinctUntilChanged) { input =>
```
mariusmuja: Why is `.distinctUntilChanged` needed here?

cornerman: Should we only do it on storageEvents?

cornerman:
I have added a version of the storage handlers without a storage event subscription, as it might useful in some cases. Not sure whether `Storage.handlerWithoutEvents` is a good name.

src/main/scala/outwatch/util/LocalStorage.scala
fdietze: Avoiding duplicate events in general. Since a change to state is not important. For example: key has value "foo" and this is set again by the handler or by a storage event from another browser window.

fdietze:
I'm fine with the name

PR #159, opened 2018-01-17
Simplify node version declaration in travis

fdietze:


LukaJCB:
Don't we want to keep using yarn? I don't really mind, but CI slowdowns wouldn't be that nice.

fdietze:
yarn is installed by default on travis.

LukaJCB:
Ooh nice! LGTM then :)

Issue #155, opened 2018-01-15
Cannot compile Seq[String]


GRBurst:
Hey there,

similar to #153, writing the following does not compile (in contradiction to my expectation):

```
div(
  Seq("first", "second")
)
```

Error:
```
[error]      type mismatch;
[error]      Clippy advises, pay attention to the marked parts:
[error]      found   : Seq[String]
[error]      required: outwatch.dom.VDomModifier
[error]      Expanded types:
[error]      found   : Seq[String]
[error]      required: cats.effect.IO[outwatch.dom.VDomModifier_]"
[error]      L112:      Seq("out", "watch"),
[error]      L112:         ^

```

mariusmuja:
Use:
```scala
div (
  Seq[VdomModifier]("first", "second")
)
```

GRBurst:
Ah yes thanks - that will do

cornerman:
Closing for now, feel free to reopen if this does not work for you.

fdietze:
Actually, since #154 we are accepting `Observable[Seq[String]]` as children. Not allowing `Seq[String]` seems to be an inconsistent API to me.

PR #160, opened 2018-01-17
Refactor implicit conversions to VDomModifier

mariusmuja:
More consistent handling of implicit conversions to `VDomModifier` though the `AsVDomModifier` type class. This removes all (except one) implicit conversions to `VDomModifier` from the `dom` package object.

It also allows to have `Seq[String]`, `Seq[Int]` (any `Seq[T : AsVomModifier]`) in the VDom (see #155).


src/test/scala/outwatch/OutWatchDomSpec.scala
```diff
@@ -824,6 +824,29 @@ class OutWatchDomSpec extends JSDomSpec {
...
+    val node = document.createElement("div")
+
+    OutWatch.renderReplace(node, div("one")).unsafeRunSync()
+    node.innerHTML shouldBe "one"
+
+    OutWatch.renderReplace(node, div(1)).unsafeRunSync()
+    node.innerHTML shouldBe "1"
+
+    OutWatch.renderReplace(node, div(1.0)).unsafeRunSync()
+    node.innerHTML shouldBe "1"
```
fdietze: Why is a `Double` `1.0` rendered as `1` and not `1.0`?

mariusmuja: Apparently that's how Scala.js renders `Double.toString`:

https://scalafiddle.io/sf/OcJkymj/0

fdietze: Ok! I just tried in the scala REPL and it was "1.0".

mariusmuja: Scala REPL is JVM.

fdietze: I wanted to say that. 😄

PR #162, opened 2018-01-21
Remove stringmodifier (work in progress)

fdietze:
I found that we could probably remove `StringModifier`, since it is converted to `StringVNode` anyways. I'm not sure If I'm missing something.

based on #161


mariusmuja:
The reason why `StringModifier` was introduced was to fix the case where something like `span("foo")` was translated in Snabbdom to `{'sel': 'span', children: [{'text': 'foo'}]}` instead of `{'sel': 'span', 'text': 'foo'}`.

Related: #117

fdietze:
I understand, thank you for explaining.

Issue #81, opened 2017-11-07
Wrap Sink.create in IO!


fdietze:
I noticed that Sink.create() does not return an IO. Is this on purpose?

cornerman:
Furthemore, we need cleanup up some remainings of the IO port. There are still [some `unsafeRunSync()` left in the code](https://github.com/OutWatch/outwatch/search?utf8=%E2%9C%93&q=unsafeRunSync+path%3Asrc%2Fmain%2Fscala&type=).

cornerman:
What about this?

cornerman:
@LukaJCB @mariusmuja can one of you enlighten us? :)

LukaJCB:
Yeah it should probably be in IO :)

PR #164, opened 2018-01-22
No need for Sink.create arguments functions to return IO

mariusmuja:
There is no need for `Sink.create` arguments functions to return `IO`s. Any side-effects that these functions might perform are suspended in the closures.

### DIFF
--- a/src/main/scala/outwatch/SideEffects.scala
+++ b/src/main/scala/outwatch/SideEffects.scala
@@ -1,11 +1,10 @@
 package outwatch
 
-import monix.execution.Scheduler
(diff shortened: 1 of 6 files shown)


mariusmuja:
This PR doesn't do anything with the return of `Sink.create` (doesn't try to fix #81), it only changes the type of the arguments  of `Sink.create` (`next`, `error` and `complete`). I don't see a reason why these argument functions need to return `IO` types (even if they return `IO` types, the returned `IO`s are evaluated immediately).

fdietze:
That makes sense, thanks for the explanation.

Issue #168, opened 2018-01-26
Use package outwatch, instead of outwatch.dom


fdietze:
Importing `outwatch.dom` can be confused with `org.scalajs.dom`. (https://gitter.im/OutWatch/Lobby?at=5a68802ee0141226508680f7)

I propose to put everything in the package `outwatch`, so the imports look like:

```scala
import outwatch._
import outwatch.dsl._
```

mariusmuja:
This seems like a good idea to me.

cornerman:
Fixed in #355

PR #178, opened 2018-03-15
Abstract over effect type WIP

### LINKED ISSUE #135: Leverage cats-effect typeclasses
Right now we're using `IO` as the single effect class, but instead we could abstract over them with the type classes found in cats-effect, i.e.

```scala
def create[F[_]: Sync, T](implicit s: Scheduler): F[Pipe[T, T]] = Sync[F].sync {
    val subjectSink = SubjectSink[T]()
    Pipe(subjectSink, subjectSink)
}
```

LukaJCB:
Similar to #144, still tons of work to be done.

fixes #135


LukaJCB:
Getting some odd illegalinheritance errors, any help? :)

src/main/scala/outwatch/Sink.scala
```diff
@@ -73,19 +74,42 @@ object Sink {
...
+    * @tparam A the type parameter of the consumed elements.
     * @return a Sink that consumes elements of type T.
     */
-  def create[T](next: T => Future[Ack],
-                error: Throwable => Unit = _ => (),
-                complete: () => Unit = () => ()
-               )(implicit s: Scheduler): IO[Sink[T]] = {
-    IO{
+  def createFull[F[+_], A](next: A => F[Unit],
+                error: Throwable => F[Unit],
```
mariusmuja: Why do `next` and `complete` need to return `F[_]`?

LukaJCB:
How would you feel about combining the `outwatch.dom._` and `outwatch.dom.dsl._` into a single import?

fdietze:
Related: #168

I would prefer to have them separated to not clutter the namespace with tags where I don't need them. But we have to think about it. Where do I need outwatch features where I don't need the dsl? It appears to me, that we mostly have the outwach package for reactive programming helpers.

LukaJCB:
I agree the separation is nice, my primary concern is that I had to introduce aliases like `VNodeF` and `VDomModifierF` in the `outwatch.dom` object, because we don't have access to the underlying `F[_]` yet, whereas in `outwatch.dom.dsl` we do have that, though I introduced another alias for `VNode` in the `OutWatchDsl` trait, which should be fine for now, I think :)

Anyways, I won't have a lot of time to continue my work on this in the next couple of weeks, so if anyone wants to contine, go right ahead! :)

src/main/scala/outwatch/Sink.scala
LukaJCB: I'd argue this makes them more reusable, in a pure application your effects are going to be in some kind of `F`, so taking an `F`, is probably much easier than having to convert it somehow.

src/main/scala/outwatch/SideEffects.scala
```diff
@@ -1,10 +0,0 @@
-package outwatch
-
-import monix.execution.Ack.Continue
-import monix.execution.Scheduler
-
-trait SideEffects {
```
fdietze: Why did you remove the sideEffects helpers?

LukaJCB: I couldn't get them to compile so I deleted them for now, and then forgot to readd, as I wasn't done

fdietze:
It compiles and the tests are running, but many are failing...

fdietze:
@cornerman, @LukaJCB and I sat together to finally resolve this PR. Unfortunately this was not successful. To summarize:

We cannot port he current Outwatch to `Effect`, since `Effect` does not have `unsafeRunSync()`, only `runAsync()`. Eliminating the last calls on the master branch lead to patching the dom asynchronously, which made the tests unreliable. We need a different architecture for outwatch and a different typeclass to describe synchronous effects.

@cornerman and @LukaJCB feel free to comment more, so that the insights don't get lost.

cornerman:
We can tackle this again now, as it should be much simpler. More details in #256

PR #185, opened 2018-04-09
support svg

cornerman:
Uses current jitpack snapshot of scala-dom-types to support svg tags and attributes.

### DIFF
--- a/src/main/scala/outwatch/dom/dsl.scala
+++ b/src/main/scala/outwatch/dom/dsl.scala
@@ -3,6 +3,7 @@ package outwatch.dom
 object dsl extends Attributes with Tags with Styles {
   object tags extends Tags {
     object extra extends TagsExtra
+    object svg extends TagsSvg
   }
   object attributes extends Attributes {
     object attrs extends Attrs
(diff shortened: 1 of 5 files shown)


src/main/scala/outwatch/dom/dsl.scala
```diff
@@ -11,12 +12,14 @@ object dsl extends Attributes with Tags with Styles {
     object events extends Events
     object outwatch extends OutwatchAttributes
     object lifecycle extends OutWatchLifeCycleAttributes
+    object svg extends AttrsSvg
   }
+  object svg extends AttrsSvg with TagsSvg
```
cornerman: Not sure whether we need both `dsl.attributs.svg` and `dsl.tags.svg` as well as `dsl.svg`.

src/test/scala/outwatch/DomEventSpec.scala
```diff
@@ -467,7 +467,7 @@ class DomEventSpec extends JSDomSpec {
         doubleStream <- Handler.create[Double]
         boolStream <- Handler.create[Boolean]
         htmlElementStream <- Handler.create[html.Element]
-        svgElementTupleStream <- Handler.create[(svg.Element, svg.Element)]
```
fdietze: What is the reason to use `Element` instead of `svg.Element`?

cornerman: Oh good catch. I think this was just a name clash between `outwatch.dom.dsl.svg` and `org.scalajs.dom.svg`. So we should keep svg.Element here.

fdietze:
I think this PR can be merged now.

fdietze:
Someone else please approve. I already used outwatch with svg in a project and it works great.

PR #191, opened 2018-04-26
event actions on EmitterBuilder

cornerman:
helper methods for calling functions on event when using an EmitterBuilder. Works for: preventDefault, stopPropagation, stopImmediatePropagation

### DIFF
--- a/src/test/scala/outwatch/DomEventSpec.scala
+++ b/src/test/scala/outwatch/DomEventSpec.scala
@@ -584,4 +584,62 @@ class DomEventSpec extends JSDomSpec {
     handler.unsafeOnNext(None)
     option shouldBe Some(None)
   }
+
+  "Emitterbuilder" should "preventDefault (compile only)" in {
+    val node = div(
+      id := "click",
+      onClick.filter(_ => true).preventDefault.map(_ => 4) --> sideEffect{()},
+      onClick.preventDefault.map(_ => 3) --> sideEffect{()}
+    )
+
+    OutWatch.renderInto("#app", node).unsafeRunSync()
+
+    val event = document.createEvent("Events")
+    initEvent(event)("click", canBubbleArg = true, cancelableArg = false)
+    document.getElementById("click").dispatchEvent(event)
+  }
+
+  "Emitterbuilder" should "stopPropagation" in {
+    var triggeredFirst = false
+    var triggeredSecond = false
+    val node = div(
+      onClick --> sideEffect{triggeredSecond = true},
+      div(
+        id := "click",
+        onClick.stopPropagation --> sideEffect{triggeredFirst = true}
+      )
+    )
+
+    OutWatch.renderInto("#app", node).unsafeRunSync()
+
+    val event = document.createEvent("Events")
+    initEvent(event)("click", canBubbleArg = true, cancelableArg = false)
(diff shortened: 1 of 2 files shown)


mariusmuja:
Looks good. But, it could be simplified like this:
```scala
  implicit class EventActions[E <: Event, O <: Event, R](builder: EmitterBuilder[E, O, R]) {
    def preventDefault: EmitterBuilder[E, O, R] = builder.map{e => e.preventDefault; e}
    def stopPropagation: EmitterBuilder[E, O, R] = builder.map{e => e.stopPropagation; e}
    def stopImmediatePropagation: EmitterBuilder[E, O, R] = builder.map{e => e.stopImmediatePropagation; e }
  }
```

This way the `eventAction` method on event builder is not needed, however the actions would need to be specified before the event is transformed (`onClick.preventDefault.map(_ => 3)` would work, but `onClick.map(_ => 3).preventDefault` would not). Is being able to place the event action after the event is transformed important to you?

cornerman:
> Is being able to place the event action after the event is transformed important to you?

Thanks for looking into it! I made that way to be able to do it, but it is not that important to me.

It might come in handy with collect, for example:
```scala
onKeyDown.collect { case e if e.keyCode == KeyCode.Enter => e.something }.preventDefault
```

which afaik would be different from putting the preventDefault in front of the collect, because unfiltered events would als be preventDefault otherwise.

mariusmuja:
Actually currently the effect should be the same, regardless where `preventDefault` is placed, because `eventAction` is applied at the beginning of the transformation chain.

```scala
def eventAction(f: E => Unit): EmitterBuilder[E, O, R] = copy(
    transformer = o => transformer(o.doOnNext(f))
 )
```

Your collect example is actually a good reason to use the `.map{e => e.preventDefault; e}` approach, in which case this would work, applying `.preventDefault` only to the filtered events:
```scala
onKeyDown.filter(_.keyCode == KeyCode.Enter).preventDefault.map(_.something)
```

cornerman:
Good point :)

Then I agree, we should use `map` for the event action. Will update this PR later.

cornerman:
@mariusmuja simplified as discussed!

src/test/scala/outwatch/DomEventSpec.scala
```diff
@@ -584,4 +584,62 @@ class DomEventSpec extends JSDomSpec {
...
+    document.getElementById("click").dispatchEvent(event)
+
+    triggeredFirst shouldBe true
+    triggeredSecond shouldBe false
+  }
+
+  "Emitterbuilder" should "stopImmediatePropagation" in {
+    // stopImmediatePropagation is supported in jsdom since version 9.12
+    // https://github.com/jsdom/jsdom/blob/master/Changelog.md#9120
+    pending
```
LukaJCB: What does this do?

fdietze: https://developer.mozilla.org/en-US/docs/Web/API/Event/stopImmediatePropagation

fdietze: Or are you talking about `pending`?

fdietze: It skips the test:
https://travis-ci.org/OutWatch/outwatch/jobs/381434109#L707

Issue #189, opened 2018-04-26
Call event.stopPropagation by default in event handlers


fdietze:
Most of the time this is confusing:

```scala
div(
  onClick --> ...,
  div(onClick --> ...)
)
```

Currently, clicking on the inner `div` also triggers the outer `onClick` handler.

cornerman:
Agree that this is confusing default-behaviour. How could we write the exception when you want it to propagate? Maybe `onClick.withPropagation`?

fdietze:
Yes, I like that.

raquo:
I agree that this would be convenient more often than not, but it would
also be a pretty significant departure from expected DOM behaviour.

Do you know if any other DOM UI library, in JS or Scala.js is doing this?

fdietze:
I searched a bit and found this: https://css-tricks.com/dangers-stopping-event-propagation/

In none of my past projects I relied on event bubbling. But I can understand that some libraries do.

I initially proposed stopping event propagation because it solves #193:

When modifying the dom from inside an event handler, the current event can bubble into the  newly created dom and trigger handlers which have nothing to do with the current event. I had exactly this issue and it was very very hard to debug. It can happen easily when switching a view based on an Observable[Boolean], where the view contains click handlers which write into the observable. I think this is a common pattern.

fdietze:
The problem is more fundamental: https://jsfiddle.net/3bjqyx2g/

I find this a bit crazy, that it works like that.

cornerman:
Good example, just updated it to be more readable: https://jsfiddle.net/3bjqyx2g/1/

raquo:
That jsfiddle example does not surprise me at all, that's how I expect things to work in JS DOM.

#193 does confuse me because it _seems_ like the parent onClick is defined on a element that doesn't exist at the time of the event, but because of a mismatch between virtual and real elements this assumption is not correct.

I don't have a strong opinion on whether Outwatch should stopPropagation by default. In general I would be leaning against doing that, but that virtualdom issue does seem worth addressing or at least documenting.

fdietze:
I'm wondering if other popular vdom libraries prevent issues like this one.

fdietze:
I tried to reproduce with React: https://codepen.io/anon/pen/OZaeyv?editors=0010

I think React does not have this problem, since it starts the rendering asynchronously.

Maybe related: https://reactjs.org/docs/faq-state.html#when-is-setstate-asynchronous

cornerman:
[...]

PR #193, opened 2018-04-26
WIP: synchronous patch in handler bubbles events into just rendered dom

fdietze:
fixes: #189


mariusmuja:
This seems to only happen when using a TrampolineScheduler with SynchronousExecution as we do in tests, I cannot replicate the behaviour with the default scheduler. An asynchronous boundary is forced before patching when merging the source observables: https://github.com/OutWatch/outwatch/blob/master/src/main/scala/outwatch/dom/helpers/DomUtils.scala#L177

cornerman:
@fdietze Do we have a test for this to check whether this is still happening?

cornerman:
> The test is in this PR

Ah thanks, I oversaw that. Will add the test to outwatch itself - and leave it pending if it is still broken. So we can fix it later.

PR #197, opened 2018-05-20
Snabbdom thunk method

### LINKED ISSUE #104: Use snabbdom thunk method
Besides just creating a new `VNodeProxy` via `h` function, we could use [thunks](https://www.npmjs.com/package/snabbdom#thunks) in snabbdom. This is a performance strategy, where the proxy is only created when the arguments it depends on change.

So, the diffing and proxy creation is saved, when we can provide a set of arguments, which sufficiently define the rendered element. Maybe we can support this?

cornerman:
Introduce `thunk` method for more efficient diffing on the relevant argument instread of comparing the `VNodeProxy`.

This PR introduces a method `.thunk` on a `VNode`, so you can do something like this:
```scala
    val myString: Handler[String] = Handler.create().unsafeRunSync()
    val renderFn: String => VDomModifier = { str =>
      Seq[VDomModifier](cls := "b", str)
    }
    val node = div(
      b(id := "bla").thunk(renderFn, myString),
      b("something else")
    )

```

Here `renderFn` will only be called if the argument from `myString` Handler changes.

fixes #104
based on #196

### DIFF
--- a/src/main/scala/outwatch/dom/Implicits.scala
+++ b/src/main/scala/outwatch/dom/Implicits.scala
@@ -5,6 +5,8 @@ import cats.syntax.apply._
 import com.raquo.domtypes.generic.keys
 import outwatch.AsVDomModifier
 import outwatch.dom.helpers.BasicStyleBuilder
+import monix.reactive.Observable
+import monix.execution.Scheduler
 
 trait Implicits {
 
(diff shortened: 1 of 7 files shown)


cornerman:
Be aware: snabbdom's `thunk` seems to rely on reference equality. I have added a pending test to verify this.

Though, there is an open issue to provide a custom comparator: https://github.com/snabbdom/snabbdom/issues/143 and https://github.com/snabbdom/snabbdom/pull/233.

cornerman:
Now, I have added a `thunk` implementation in scala which respects equality properly.

cornerman:
Thinking about this feature...is it really different from just ignoring duplicates in the observable and thereby recreating the `VNodeProxy` only if the streamed values change?

mariusmuja:
> Thinking about this feature...is it really different from just ignoring duplicates in the observable and thereby recreating the VNodeProxy only if the streamed values change?

I was just going to ask this, I don't see why `thunk` is any better then just `div(observable.distinctUntilChanged)`. Plus .`distinctUnilChanged` uses `cats`'s `Eq` typeclass.

cornerman:
Exactly! So, I am just going to close this :)

PR #199, opened 2018-05-21
stream receiver for Observable[VDomModifier]

### LINKED ISSUE #198: StreamReceiver for VDomModifier
We have an `AttributeStreamReceiver` for streaming a single Attribute with a fixed name and `Child-/ChildrenStreamReceiver` for `VNode`.

Can we have a possibility to render an `Observable[VDomModifier]` as more generic solution?

cornerman:
Added a Modifier for supporting an `Observable[VDomModifier]`: `ModifierStreamReceiver`.

This is support stream arbitrary modfiers and not just nodes and single attributes. The `ChildStreamReceiver` and `ChildrenStreamReceiver` are replaced with the ModifierStreamReceiver. Furthermore, the StaticNodeRender was removed. The AttributeStreamReceiver still makes sense, because it can be made unique per attribute title.

Questions:
* How about streaming a `Key`? This does not really make sense with snabbdom or does it?

fixes #198 #187


mariusmuja:
I didn't have time to look over this yet, but after a quick glance I'm wondering if we want the library to handle nested Observables. 

I think it might be better to not do so and require the client code to flatten the observables before passing them to outwatch. For example, this PR uses `switchMap` to flatten nested observables, but why not use `mergeMap` (or `concatMap`). Which is the correct behaviour? Probably depends on the application.

mariusmuja:
> For the usage, I find this really convenient: I can just combine VDomModifiers whether they are streamed or not. Additionally I can have a Seq of modifiers streamed into a node which leads to less updates than having multiple attribute streams. 

I like the idea of supporting Seq of modifiers streamed into a node and I see your point on `switchMap` having the same behaviour as static nodes with children observables being replaced. 

What I'm not so convinced about yet is the usefulness of having observables of VDomModifiers streamed into a node.

cornerman:
> What I'm not so convinced about yet is the usefulness of having observables of VDomModifiers streamed into a node.

I like that I can treat all modifiers equally, often I need to make one part reactive, and now this leads to nesting of elements.

Another advantage is that I can better separate reactive components and behaviours. Say, I am having a simple component with a handler:

```scala
def component = Handler.create[String].flatMap { handler =>
   span("Like: ", handler, onClick("yes!") --> handler)
}

// usage
div(component)
```

This is nice and reuseable, but what if I want control over the outer element? What I want is to render this behaviour into an element:

```scala
def component = Handler.create[String].flatMap { handler =>
   VDomModifier(handler, onClick("yes!") --> handler)
}

// usage
div("Love: ", component)
```

It could even include more observables and would still work. Of course, one should watch out what you are combining but that is already the case now.

mariusmuja:
Yeah, that's nice! But this example doesn't use higher-order observables. Did you encounter instances where using higher order observables was needed?

I agree that it's nice to treat all VDomModifies equally, but maybe a solution would be to adjust the Modifier hierarchy? I'm a bit reluctant to introducing this complexity without a well defined use case.

cornerman:
> but maybe a solution would be to adjust the Modifier hierarchy?

Indeed, we could adjust the Modifier hierarchy, maybe with a `StaticModifier`. But I would like to see whether we can come up with something that works for all modifiers :)

I extended the component in the example to have a handler of VDomModifier. Does that make sense?

mariusmuja:
> I extended the component in the example to have a handler of VDomModifier. Does that make sense?

Yes, but that still doesn't use higher order observables.

cornerman:
> Yes, but that still doesn't use higher order observables.

If I wrap the usage in an observable, I would have nested streamed modifiers. I thought that was a higher order observable - isn't it?

mariusmuja:
> If I wrap the usage in an observable, I would have nested streamed modifiers. I thought that was a higher order observable - isn't it?

Yes, it's obviously possible to construct an example of usage. What I'm wondering about is a practical use case for this, when having nested observables of VDomModifiers part of a node is the most obvious  or the easiest / most intuitive way of solving the problem at hand.

fdietze:
What I like about outwatch, is that you can throw almost everything at it, and it will render it. This PR makes this even more flexible. Even if it's hard to find an artificial use-case of nested observables right now, we do not know the architectures other people have.

PR #203, opened 2018-07-12
Remove Sink and Handler

fdietze:
In our code base we recognized that we are working
 a lot with `Observable` and `Observer`/`Subject` from Monix. It is cumbersome to always wrap these in `Handler` and `Sink` just to use them with outwatch.

I think many people already know the Monix Api and should not need to learn a new one with the same functionality.

What do you think?

Please provide feedback especially for `MonixOps`: https://github.com/OutWatch/outwatch/pull/203/files#diff-99d5d9668fb74825f242be5dccf96340


src/test/scala/outwatch/SnabbdomSpec.scala
```diff
@@ -30,7 +30,7 @@ class SnabbdomSpec extends JSDomSpec {
     import outwatch.dom._
     import outwatch.dom.dsl._
 
-    val clicks = Handler.create[Int](1).unsafeRunSync()
+    val clicks = outwatch.Handler.create[Int](1).unsafeRunSync()
```
cornerman: should not be needed anymore

src/main/scala/outwatch/util/LocalStorage.scala
```diff
@@ -42,18 +41,21 @@ class Storage(domStorage: dom.Storage) {
         None
     }
 
-  def handlerWithoutEvents(key: String)(implicit scheduler: Scheduler): IO[Handler[Option[String]]] = {
-    handlerWithTransform(key, identity)
+  def subjectWithoutEvents(key: String)(implicit scheduler: Scheduler): IO[Subject[Option[String], Option[String]]] = {
+    subjectWithTransform(key, identity)
   }
 
-  def handlerWithEventsOnly(key: String)(implicit scheduler: Scheduler): IO[Handler[Option[String]]] = {
```
cornerman: deprecated handlerWithEventsOnly

src/main/scala/outwatch/util/LocalStorage.scala
```diff
@@ -42,18 +41,21 @@ class Storage(domStorage: dom.Storage) {
         None
     }
 
-  def handlerWithoutEvents(key: String)(implicit scheduler: Scheduler): IO[Handler[Option[String]]] = {
```
cornerman: deprecated handlerWithoutEvents

mariusmuja:
> @mariusmuja what do you think of this PR? Does it break a lot of code for you, because it removes Pipe?

I use the `Pipe`/`PipeOps` almost exclusively in `outwatch-extras` and all my `outwatch` code depends on `outwatch-extras`, so yes, kind of. On the plus side I only need to update `outwatch-extras`.

fdietze:
@mariusmuja I just brought back the operators and made the naming more Monix compatible, so the transition should be easier.

PR #204, opened 2018-07-13
fix nested VDomModifier-Observable with CompositeModifier

cornerman:
composite modifiers were not handled correctly when they are nested. This lead to duplicate static nodes being rendered.


cornerman:
Please do not merge, we have found another problem, see branch `vdommodifier-bug2`.

cornerman:
We have now refactored the `Receivers` in this fix. Composite modifiers are now handled with the same mechanism as the Receivers. This fixes the problem in the test we have added.

src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -135,61 +137,101 @@ private[outwatch] final case class SeparatedEmitters(
...
+private[outwatch] class StreamableModifiers(modifiers: Seq[Modifier]) {
+
+  //TODO: hidden signature of this method (we need StaticModifier as a type)
+  //handleStreamedModifier: Modifier => Either[StaticModifier, Observable[StaticModifier]]
+  private val handleStreamedModifier: Modifier => ContentKind = {
+    case ModifierStreamReceiver(modStream) =>
+      val observable = modStream.switchMap[Modifier] { mod =>
+        handleStreamedModifier(mod.unsafeRunSync) match {
+          case ContentKind.Dynamic(stream) => stream.startWith(EmptyModifier :: Nil)
+          case ContentKind.Static(mod) => Observable(mod)
```
LukaJCB: Use `Observable.now`

PR #222, opened 2018-08-10
Modifier stream with value

cornerman:
This PR has become quite the collection of changes which we did when working on our application. When investigating performance issues, we saw that the number of snabbdom patches needed to render our initial page was really high (multiple hundred times depending on the number of items displayed). Furthermore, updates on small subcomponents were triggering multiple times.

When using a stream of VDomModifier, outwatch would always render the element with an `empty placeholder` first, and then apply updates in form of snabbdom patches. In our case, we always have a `current value` (because our state is modeled with scalarx, same holds if you are using monix Var). Therefore, we know that we are able to render our initial page without any patching, because we know the initial state.

This lead to the first part of this PR: Introducing an initial value for `ModifierStreamReceiver`. We can thereby render initially with a sane value if it is available. Otherwise, we can still default to `EmptyModifier` as we do now (when using an Observable without any initial value). The initial patching can thereby be circumvented.

Down the line, we saw that calculating the VNodeProxy was quite compute heavy in js and that we do it way too often. When an element has multiple streams, each stream can trigger an update. In the case that one stream triggers, outwatch would calculate the VNodeProxy of all of its children. So when multiple streams trigger, all children are reiterate on each of these triggers. This is why, we introduced a state in the VNode (it caches the VNodeProxy, so it can be reused efficiently). Caching is always ugly, but in this case it is quite safe: each VNode is only ever used once, because it is wrapped in IO.

Now, the cache enables us to fix another problem. Let's say we have something like this: `div ( numberObservable.map(number => span(number)) )`. When the div initialyl renders, it is empty, because there is no initial value. When the observable then triggers, it will update the div to contain some number. Everything fine. But now lets say, the div itself was also part of a stream. If the number observable changes, this is only known to the div. But no parent knows of these changes. Therefore, the initial render of the outer div will remove the inner span from the dom. And then the number observable will trigger, and fill the dom again. This is really a problem when working with virtual dom, because we lose all benefits of patching. We just remove and add the node again.

Therefore, we have added a very ugly mutation. On each update, the just introduced cache of the VNode is modified to carry the new value. Therefore every parent's cache will now also contain the updated value because it contains the same VNodeProxy. So rendering will not remove the things from dom which should actually still be there.

On the way, we have also removed `StringModifer`, which we know handle as a StringVNode. But it still behaves as before so that if there are only strings, only a string is send to snabbdom. Furthermore, the `AttributeStreamReceiver` was removed and replaced by the `ModifierStreamReceiver`. The only benefit of the AttributeStreamReceiver was that it could make the stream unique per element. But we cannot really benefit from this behaviour because it hides a problem from the user (double definition) and a normal ModifierStreamReceiver could still lead to duplicate attribute definitions.

Sorry for the long post and the many changes. But this was a longer effort and the changes are somehow intermingled and partly depend on each other. What do you think?

### DIFF
--- a/src/main/scala/outwatch/dom/helpers/Snabbdom.scala
+++ b/src/main/scala/outwatch/dom/helpers/Snabbdom.scala
@@ -258,7 +258,7 @@ private[outwatch] trait SnabbdomModifiers { self: SeparatedModifiers =>
     // never contain streamable content and therefore we do not need to handle
     // them with Receivers.
     val (receivers, dataObject) = previousProxy.fold {
-      val receivers = Receivers(childrenWithKey, attributeReceivers)
+      val receivers = Receivers(childrenWithKey)
       (Option(receivers), createDataObject(receivers))
     } { p =>
       (None, updateDataObject(p.data))
--- a/src/main/scala/outwatch/dom/helpers/Snabbdom.scala
+++ b/src/main/scala/outwatch/dom/helpers/Snabbdom.scala
@@ -264,14 +264,16 @@ private[outwatch] trait SnabbdomModifiers { self: SeparatedModifiers =>
       (None, updateDataObject(p.data))
     }
 
-    val initialProxy = childrenWithKey match {
-      case Children.VNodes(vnodes, _) =>
-        val childProxies: js.Array[VNodeProxy] = vnodes.collect { case s: StaticVNode => s.toSnabbdom }(breakOut)
-        hFunction(nodeType, dataObject, childProxies)
-      case Children.StringModifiers(textChildren) =>
-        hFunction(nodeType, dataObject, textChildren.map(_.string).mkString)
-      case Children.Empty =>
+    val initialProxy = {
+      if (childrenWithKey.nodes.isEmpty) {
         hFunction(nodeType, dataObject)
+      } else if (childrenWithKey.hasVTree) {
+        val childProxies: js.Array[VNodeProxy] = childrenWithKey.nodes.collect { case s: StaticVNode => s.toSnabbdom }(breakOut)
+        hFunction(nodeType, dataObject, childProxies)
+      } else {
+        val textChildren: js.Array[String] = childrenWithKey.nodes.collect { case s: StringVNode => s.string }(breakOut)
+        hFunction(nodeType, dataObject, textChildren.mkString)
+      }
     }
 
     // we directly update this dataobject with default values from the receivers
--- a/src/main/scala/outwatch/dom/helpers/Snabbdom.scala
+++ b/src/main/scala/outwatch/dom/helpers/Snabbdom.scala
@@ -3,8 +3,8 @@ package outwatch.dom.helpers
 import monix.reactive.subjects.PublishSubject
 import monix.execution.Ack.Continue
 import monix.execution.Scheduler
-import monix.execution.cancelables.SingleAssignCancelable
 import monix.reactive.Observable
+import monix.execution.cancelables.StackedCancelable
 import org.scalajs.dom
 import outwatch.dom._
 import snabbdom._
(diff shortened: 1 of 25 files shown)


src/main/scala/outwatch/dom/helpers/Snabbdom.scala
```diff
@@ -257,11 +237,17 @@ private[outwatch] trait SnabbdomModifiers { self: SeparatedModifiers =>
   // dynamic modifier of this node yields a new VNodeState, this new state with
   // new modifiers and attributes needs to be applied to the current
   // VNodeProxy.
-  private def toProxy(nodeType: String, previousProxy: Option[VNodeProxy])(implicit scheduler: Scheduler): VNodeProxy = {
+  def toProxy(nodeType: String, previousProxy: Option[VNodeProxy])(implicit scheduler: Scheduler): VNodeProxy = {
+
+    // previousProxy match {
```
fdietze: please remove commented code.

src/main/scala/outwatch/dom/helpers/Snabbdom.scala
```diff
@@ -257,11 +237,17 @@ private[outwatch] trait SnabbdomModifiers { self: SeparatedModifiers =>
...
+    //   case Some(p) => org.scalajs.dom.console.log("PATCHING", p)
+    //   case None => println("NEW NODE")
+    // }
 
     // if child streams exist, we want the static children in the same node to have keys
     // for efficient patching when the streams change
     val childrenWithKey = children.ensureKey
+    // println("TOPROXY WITH CHILDREN " + children)
```
fdietze: remove

src/main/scala/outwatch/dom/helpers/Snabbdom.scala
```diff
@@ -287,15 +273,53 @@ private[outwatch] trait SnabbdomModifiers { self: SeparatedModifiers =>
...
+    }
+
+    proxy
   }
 
-  private[outwatch] def updateSnabbdom(previousProxy: VNodeProxy)(implicit scheduler: Scheduler): VNodeProxy = toProxy(previousProxy.sel, Some(previousProxy))
-  private[outwatch] def toSnabbdom(nodeType: String)(implicit scheduler: Scheduler): VNodeProxy = toProxy(nodeType, None)
+  private[outwatch] def updateSnabbdom(previousProxy: VNodeProxy)(implicit scheduler: Scheduler): VNodeProxy = {
+    val  p = toProxy(previousProxy.sel, Some(previousProxy))
+    // org.scalajs.dom.console.log("PROXY " , JSON.stringify(p, space = 2))
```
fdietze: remove

src/main/scala/snabbdom/hFunction.scala
```diff
@@ -33,10 +33,10 @@ object hFunction {
 
 
 trait Hooks extends js.Object {
-  val insert: js.UndefOr[Hooks.HookSingleFn]
+  var insert: js.UndefOr[Hooks.HookSingleFn] //TODO
```
fdietze: TODO what?

fdietze:
I just found another bug and added a failing test case. :wink:

cornerman:
Good find, this because we are using a `PublishSubject` with `Handler`. If you replace the `Handler` with a `BehaviorSubject` or make `replayLimited(1)`, the test is fixed.

This happens because there are now less `patch`-calls when nesting observables. Therefore it seems to be important to have the last value in the Observable.

How about we make `Handler.apply(value)` a BehaviorSubject and `Handler.apply` a limited `ReplaySubject`?

fdietze:
> How about we make Handler.apply(value) a BehaviorSubject and Handler.apply a limited ReplaySubject?

I agree. I think this reflects more the intent of the developer.

cornerman:
I have changed this on a branch (https://github.com/cornerman/outwatch/tree/replay-behaviour-subject). But now there are two other failing tests. Not sure whether this is right. Can you have a look? :)

cornerman:
The test you have added fails on master, too!

cornerman:
Ok, I have fixed the other test, too. I think this is correct now!

@LukaJCB What do you think?

fdietze:
Great work @cornerman!

I think you can merge now, since the bugs are fixed, it is a real improvement and it even fixes bugs currently present on the master branch.

PR #226, opened 2018-08-28
use div instead of app as tag for Outwatch.renderInto

### LINKED ISSUE #223: Outwatch.renderInto() uses a common magic value
I had difficulty using `Outwatch.renderInto` so I eventually hacked a solution using `renderReplace`.  It turns out I was using an element id `"app"` and renderInto was creating one with the same id.

Perhaps the magic value "app" should get changed to something less likely to collide in the wild? 

```
  def renderInto(element: dom.Element, vNode: VNode)(implicit s: Scheduler): IO[Unit] = for {
    node <- vNode
    _ <- IO {
      val elem = dom.document.createElement("app")
      element.appendChild(elem)
      patch(elem, node.toSnabbdom)
    }
  } yield ()
```

cornerman:
fixes #223

### DIFF
--- a/src/main/scala/outwatch/dom/OutWatch.scala
+++ b/src/main/scala/outwatch/dom/OutWatch.scala
@@ -12,7 +12,7 @@ object OutWatch {
   def renderInto(element: dom.Element, vNode: VNode)(implicit s: Scheduler): IO[Unit] = for {
     node <- vNode
     _ <- IO {
-      val elem = dom.document.createElement("app")
+      val elem = dom.document.createElement("div")
       element.appendChild(elem)
       patch(elem, node.toSnabbdom)
     }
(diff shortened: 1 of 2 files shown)


LukaJCB:
This test seems to fail:

```
should support emitter operations *** FAILED ***
```

cornerman:
Interesting. This is because we are now patching one node with the same `sel` (both div). And now `patch` is called instead of `insert` by snabbdom. Before we were patching a tag `app` with a different tag `div`.

cornerman:
I am working on new hooks `onDomMount`, `onDomUpdate`, `onDomUnmount`. These can reliably tell whether a node is newly inserted/patched or is just updated. So when using these hooks, we can fix the tests.

cornerman:
Fixed the tests for now by ensuring keys on the relevant inserted nodes.

Issue #217, opened 2018-07-27
Merging emitter functions prevents stopImmediatePropagation


fdietze:
https://github.com/OutWatch/outwatch/blob/b6e6d8600686c0e1af729ae42d30d885cb3bd049/src/main/scala/outwatch/dom/helpers/Snabbdom.scala#L174

cornerman:
We need to send a list of emitter functions (in order) to snabbdom instead of just one.

cornerman:
This will never work with snabbdom, because snabbdom creates one(!) eventlistener for each occuring event type.

So, my best guess would be to remove stopImmediatePropagation from our API, as we cannot make this work. You can still implement this functionality in your own application.

cornerman:
As long as we are using snabbdom as is, we cannot fix it.

PR #231, opened 2018-09-09
remove stopImmediatePropagation from API (see #217)

cornerman:
Snabbdom registers only one event listener for each occurring `eventType` in a VNode. This is why, we cannot implement #217 with snabbdom. Multiple handlers, will still yield only one event listener. Then, the dom cannot stop immediate propagation on these handlers.


cornerman:
For future readers: If we really want this feature, we need to have our custom snabbdom eventlistener-module and implement it differently. But I am not sure, whether this is worth it. We need to be careful when removing event listener in a patch.

cornerman:
any thoughts?

PR #232, opened 2018-09-09
tests: remove unsafeRunSync usage

ahjohannessen:
- add `AsyncFlatSpec` from scalatest and introduce
   two variants of `JSDomSpec`.

 - refactor tests that rely on `unsafeRunSync` to
   use `unsafeToFuture` from scalatest. Furthermore,
   there is an implicit conversion from `IO[Assertion]`
   to `Future[Assertion]` in `JSDomAsyncSpec` that reduces
   usage of `unsafeToFuture` boilerplate.

### DIFF
--- a/src/test/scala/outwatch/OutwatchSpec.scala
+++ b/src/test/scala/outwatch/OutwatchSpec.scala
@@ -1,6 +1,8 @@
 package outwatch
 
-
+import scala.language.implicitConversions
+import scala.concurrent.Future
+import cats.effect.IO
 import monix.execution.Ack.Continue
 import monix.execution.ExecutionModel.SynchronousExecution
 import monix.execution.schedulers.TrampolineScheduler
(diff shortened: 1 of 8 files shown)


src/test/scala/outwatch/OutwatchSpec.scala
```diff
@@ -1,20 +1,22 @@
...
 import org.scalatest.{BeforeAndAfterEach, FlatSpec, Matchers}
+import org.scalatest._
 import outwatch.Deprecated.IgnoreWarnings.initEvent
 
-
 trait EasySubscribe {
 
   implicit class Subscriber[T](obs: Observable[T]) {
-    def apply(next: T => Unit)(implicit s: Scheduler): Cancelable = obs.subscribe { t =>
+    def apply(next: T ⇒ Unit)(implicit s: Scheduler): Cancelable = obs.subscribe { t ⇒
```
fdietze: I'd prefer to stay with ascii arrows instead of unicode ones.

ahjohannessen: I'll update it to use ascii

ahjohannessen: @fdietze updated PR to use `<-` instead of `←` and `=>` instead of `⇒` in tests

fdietze:
LGTM! Thank you for your contribution!

PR #237, opened 2018-09-11
add element type parameter to VTree and introduce alias VNodeT

cornerman:


src/main/scala/outwatch/dom/package.scala
```diff
@@ -1,13 +1,15 @@
...
 import cats.effect.IO
+import org.scalajs.dom.Element
 
 package object dom extends Implicits with ManagedSubscriptions with SideEffects with MonixOps {
 
   val Sink = outwatch.Sink
   val Handler = outwatch.Handler
 
-  type VNode = IO[VTree]
+  type VNodeT[+T <: Element] = IO[VTree[T]]
```
LukaJCB: I don't think I like the `VNodeT` name. It uses the same convention as Monad transformers in cats, which might confuse done people familiar with them 😁

cornerman: True, that name is not really intuitive. Do you have a better name? :)

cornerman:
There is one minor annoyance, you cannot write this anymore:
```scala
val handler: Handler[VNode] = Handler.create(div("hallo"))
```

Because Handler is invariant in its type param, we need write one of these two versions:
```scala
val handler: Handler[VNode] = Handler.create[VNode](div("hallo"))
// or
val handler: Handler[VNodeT[html.Div]] = Handler.create(div("hallo"))
```

src/main/scala/outwatch/dom/package.scala
cornerman: @LukaJCB @fdietze What would be a good name for `VNodeT`?

fdietze: `TypedVNode`?

fdietze: `GenericVNode`?

fdietze:
> There is one minor annoyance, you cannot write this anymore:
> 
> ```scala
> val handler: Handler[VNode] = Handler.create(div("hallo"))
> ```

I think this use-case is very rare. It should be fine.

Issue #239, opened 2018-09-17
Add DomPreUpdateHook 


cornerman:
Similar to `onSnabbdomPrePatch`, we want to have a high-level hooks for being called before the update takes place.

Related: #228

Issue #236, opened 2018-09-10
Emitter and hooks have async boundary


cornerman:
We use observer as our sink for emitters and hooks. Thereby, we have an async boundary between dom event handler and our outwatch sink. Then things like `stopPropagation`, `preventDefault` might not work realiably. We should be sync here.

Should we change emitter to accept just functions? We anyhow only use `onNext` on the observer .

Issue #227, opened 2018-09-06
PreventDefault, StopPropagation helpers on window/document events


cornerman:
Same functionality like we have for EmitterOps.

cornerman:
We have `EmitterBuilder.fromObservable(obs)` now, which helps. But you are not synchronous but already in an observable there, which is why you do not have `prevenDefault`/`stopPropagation`.

Issue #219, opened 2018-07-30
Use lambda directly instead of sideEffect


cornerman:
Can we get rid of the `sideEffect` function for supplying a lambda instead of an observer (for event emitter and hooks)?

cornerman:
You can use `emitter.foreach(action)` now.

PR #247, opened 2018-11-04
Performance-oriented rewrite

cornerman:
This PR is pretty much a rewrite of the outwatch core (mainly Snabbdom.scala and DomUtils.scala). It was driven by bugs we have found and performance requirements when using outwach in a web app we develop.

When benchmarking this PR in a simple test against the current master, I get speed improvements by a factor of about 3-4. I see similar figures when manually testing performance in the browser.

Here is a list of relevant changes:
* Make VDomModifier non-sideeffecting by default. There is no dom-access or subscriptions happening when constructing VNodes. This means that a modifier is no longer wrapped in IO. Instead `div` just returns a `VNode` and `href := "meh"` just returns an Attribute. Only if you apply sideeffects via a handler or from your own code, you have an IO around your node: `Handler.create("text").map(div(_)): IO[VNode]`. You can now transparently embed it into your main logic and unwrap the IO for using `Outwatch.renderInto`. Otherwise you can just use `IO[VDomModifier]` in your code as before: `div(ioNode: IO[VNode], ioMod: IO[VDomModifier])`. These IO modifiers are represented as an `EffectModifier`. The whole construction is still referential transparent, it is just much simpler and less intrusive.
* Use our own fork of snabbdom. It fixes bugs that are still not merged in snabbdom and allows us to use it more efficiently. For example, we can reuse VNodeProxies to have a very intelligent way of caching VNodes in outwatch. That allows for mixing Observables without recalculating the whole subtree on every update. I think the goal needs to be to write our own library like snabbdom or something similar!
* Fix leaking subscriptions in emitter builders and monix ops
* Faster code with less memory allocations for constructing VNodeProxies, a lot of mutability but just internally :)
* Remove implicit scheduler from builder dsl, we have a `SchedulerAction` modifier now, which gets the scheduler injected from the outer `Outwatch.render` call. We delay subscriptiuons in a `SchedulerAction` until a node is going to be inserted and use the same scheduler for the whole ui of the app.
* Child commands to send an `Observable[ChildCommand]` that can incrementally update a list of nodes. Gives major improvments in speed as opposed to rerendering the whole list of nodes like a `Observable[Seq[VNode]]`.
* Add thunk nodes. This is also useful for performance reasons, but you can keep writing your components as before with `Observable[Seq[VNode]]`. You can define a thunk node like this: `div.thunk("unique-component-name")(arg1, arg2)(VDomModifier(arg1, arg2, ...))`. Here `arg1` and `arg2` are the only dependencies of this div, so it only needs to be rerendered if one of these changes. So thunk delays the execution of the VDomModifier in the last parameter list (call-by-name) and only computes a new node if one of the arguments has changed. if the arguments stay the same, it can just reuse the node from the previous run (identified by the "unique-component-name"). For the perofmance this comes close to ChildCommands. It works similar to React components with a shouldRender function, which compares the previous and the next state.
* Sync emitters so that stopPropagation/preventDefault actually work and are not behind an async boundary. Only async if needed.
* Add `mapTo` to emitter builder to get the value to be mapped by-name.
* Add `mapOption` to emitter builder to map to an option.
* Add `useLatest` to emitter builder for combining two emitter builders.
* Add `foreach` to emitter builder. foreach fits naturally with methods like `map`, `collect`, etc and gives a way to interact nicely with the dom: `onDomMount.asHtml.foreach { elem => elem.focus }`. The sideeffect factories are deprecated.
* Userful emitter builder factories: `EmitterBuilder.fromObservable(Observable)` or `emitter(Observable)` and `EmitterBuilder.ofModifier`.
* Less function allocations when handling hooks and emitters.
* Do not return IO in Sink.create. The sink itself is not a sideeffect. If there is a sideeffect and therefore a reason for IO, then it lies in the function. Now then the function should be in IO beforehand. So, you will use it this way anyhow: `statefulAction.map(Sink.create _)`.
* Add DomPreUpdate hook: `onDomPreUpdate`
* Add Init hook, but it carries not element and is more of interest when working with the `VNodeProxy`, so you can only use it wirk InitHook(f: proxy => Unit).
* Add Handler.unsafe method to get a handler without IO. This may be needed for a global state and is easier to write but shares the logic with Handler.create
* Do not enforce keys in outwatch, we are able to track the identity of a vnode via their `_id` and do not need to enfroce a key for streaming nodes. Letting the user do this on demand is much better and allows for improvements (that are not broken by automatic key generation in outwatch).
* Add `managedAction`, `managedElement`, `managedElement.asHtml`, `managedElement.asSvg` as helper for binding things to the lifetime of an element. Allows easy integration of third-party libraries
* Remove syntax suger `=?`
* Add a minimalistic performance test to get an idea of the figures.
* Remove `dsl.attributes.lifecycle._`, you can use `dsl.attributes.outwatch._` already.
* More inlining of simple redirect functions. Specialized versions of functions for `VDomModifier` to not require a `AsVDomModifier`.
* Fix streaming of dom-mount-hooks and managed subscriptions. Now you can have `Observable(managed(subscription), onDomMount --> sink)` and it will work intuitively and trigger when streamed in and out accordingly.
* Update monix to experimental RC2 and cats-effect to 1.0 
* Do not use `h`-function of snabbdom, it is slow because it does a lot of checks that we do not need and it reiterates all svg children which we also do not need because we have this information at construction time from dom-types. Therefore we just construct a new `VNodeProxy` instead of calling the h-function.
* Add VNode.prepend for prepending modifiers to a Modifier, this is useful for providing defaults to a VNode from the call-side. 
* Add `HtmlVNode` and `SvgVNode` to differentiate between html and svg nodes.

For the user, the main difference should be that `VDomModifier` is no longer sideeffecting and therefore is not wrapped in an IO anymore. A common difference:
This does not longer work: `Handler.create("text").flatMap(handler => div(handler))`. You now write : `Handler.create("text").map(handler => div(handler))`. The same happens with for comprehensions which expect to flatMap over an IO but get a VNode instead: `for { handler <- ...; node <- div(...) } yield node`. You can now write: `for { handler <- ... } yield div(...)`.

I am sorry, this is such a huge PR, but I am not really able to split these changes apart in a meaningful way. I could squash the changes into a single commit but not sure if that makes sense.

You can try it via jitpack: `"com.github.cornerman" % "outwatch" % "5fd78a6"`

Fixes: #246, #239, #236, #227, #219, #214


src/main/scala/outwatch/dom/ValueObservable.scala
```diff
@@ -3,22 +3,29 @@ package outwatch.dom
 import cats.Functor
 import monix.reactive.Observable
 
-trait ValueObservable[+T] {
+trait ValueObservable[+T] { self =>
```
fdietze: This trait is only talking about the initial value, isn't it? Took me some moments to figure that out. Should we rename it, or write a comment?

src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -1,180 +1,339 @@
...
-  private val handleStreamedModifier: Modifier => ContentKind = {
-    case ModifierStreamReceiver(modStream) =>
-      val observable = modStream.observable.switchMap[Modifier] { mod =>
-        handleStreamedModifier(mod.unsafeRunSync) match {
-          //TODO: why is startWith different and leaks a subscription? see tests with: stream.startWith(initialValue :: Nil)
-          case ContentKind.Dynamic(stream, initialValue) => Observable.concat(Observable.now(initialValue), stream)
-          case ContentKind.Static(mod) => Observable.now(mod)
-        }
-      }
+private[outwatch] class NativeModifiers(
```
fdietze: A short comment describing what `NativeModifiers` are, would be nice.

src/main/scala/snabbdom/hFunction.scala
```diff
@@ -84,41 +67,105 @@ object DataObject {
...
-      val attrs: js.Dictionary[AttrValue] = _attrs
-      val props: js.Dictionary[PropValue] = _props
-      val style: js.Dictionary[StyleValue] = _style
-      val on: js.Dictionary[js.Function1[Event, Unit]] = _on
-      val hook: Hooks = _hook
-      val key: UndefOr[KeyValue] = _key
+  def empty: DataObject = new DataObject { }
+}
+
+// @js.native
```
fdietze: Why are these comments here?

src/main/scala/snabbdom/hFunction.scala
```diff
@@ -84,41 +67,105 @@ object DataObject {
...
+// }
+
+// @js.native
+// trait thunkFunction extends js.Any {
+//   def apply(selector: String, renderFn: js.Function, argument: js.Array[Any]): VNodeProxy = js.native
+//   def apply(selector: String, key: String, renderFn: js.Function, argument: js.Array[Any]): VNodeProxy = js.native
+// }
+object thunk {
+  // own implementation of https://github.com/snabbdom/snabbdom/blob/master/src/thunk.ts
+  //does respect equality. snabbdom thunk does not: https://github.com/snabbdom/snabbdom/issues/143
```
fdietze: equality by the `equals` method?

fdietze:
```suggestion
  //does respect equality via the equals method. snabbdom thunk uses reference equality: https://github.com/snabbdom/snabbdom/issues/143
```

src/main/scala/snabbdom/hFunction.scala
fdietze: I think we should either remove them or write a short note, why we left them here.

```suggestion
// These are the original facades for snabbdom thunk. But we implement our own, so that for equality checks, the equals method is used.
// @js.native
```

src/test/scala/outwatch/PerfTest.scala
```diff
@@ -0,0 +1,191 @@
...
+import outwatch.dom._
+import outwatch.dom.dsl._
+
+import scala.scalajs.js
+
+class PerfTest extends JSDomSpec {
+
+  val numIterations = 100
+
+  "Perf" should "be" in {
```
fdietze: Benchmarks should not run when the tests run.
I suggest a sbt subproject `bench` where the benchmarks are called by the `main` method and can be run via `sbt bench/run`.

fdietze:
I read through all the code and left a few comments.
Most of the time it made sense to just look at the new file instead of the diff.

Overall, it looks very good to me and is a major improvement for Outwatch. Thank you!

src/main/scala/outwatch/dom/helpers/DomUtils.scala
cornerman: Will do :)

src/main/scala/outwatch/dom/ValueObservable.scala
cornerman: Good point. A comment would definitely help. And I am investigating whether we can have `type ProHandler = ValueObservable with Observer` instead of `Observable with Observer`. This allows for less update coming from handler. We can implement a Handler similar to the `BehaviorSubject` which tracks the last emitted value of the stream and then provides some common operators to transform a ValueObservable similar to an Observable.

Any ideas for a better name? It is an Observable with an initial/current value.

fdietze: `ObservableWithInitialValue` :wink:  ?

ngbinh:
This PR looks amazing!

src/main/scala/outwatch/dom/ValueObservable.scala
cornerman: Let's leave this part for #254 where we are anyhow touching the ValueObservable type. This type is already on master, so we can check there for renaming too.

PR #254, opened 2018-11-08
[WIP] Let Handler/ProHandler be a ValueObservable

### LINKED ISSUE #250: Use Var in Handler.create(seed)
A `monix.Var` has a current value that is synchronously available, so we have a proper ValueObservable which can render the initial state more efficiently.

Todos:
* Handler is `Observer with Observable` and Var is only `Observable` with a `.value()` method.
* Common mapping methods should return a `Var` or ValueObservable. Only then you can actually benefit in your components, because you will normally map/etc over your observable.

cornerman:
This PR adds our own `BehaviorProHandler` which implements a `ValueObservable`. It is implemented very similar to the `BehaviorSubject` in monix. But it additionally exposes the current value of the stream, so we can access it synchronously. This is really nice from a performance perspective, because we can render the initial state of a reactive component without needing to patch the first emitted value of the observable.

We have implemented some common operations like map/filter/collect/scan... on the ValueObservable so you can transform the ValueObservable without losing the initial value downstream. There is an implicit from ValueObserable => Observable, so you can use complex observable methods directly on the ValueObserable and do not have to change your code there. 

Based on #247, so only look at the last four commits. Fixes #250 (though not directly by using Var).

### DIFF
--- a/outwatch/src/main/scala/outwatch/AsVDomModifier.scala
+++ b/outwatch/src/main/scala/outwatch/AsVDomModifier.scala
@@ -80,15 +80,31 @@ object AsVDomModifier {
   implicit def effectRender[T : AsVDomModifier]: AsVDomModifier[IO[T]] = (effect: IO[T]) =>
     EffectModifier(effect.map(VDomModifier(_)))
 
+  implicit object ValueObservableRender extends AsVDomModifier[ValueObservable[VDomModifier]] {
+    @inline def asVDomModifier(value: ValueObservable[VDomModifier]): VDomModifier = ModifierStreamReceiver(value)
+  }
+
   implicit def valueObservableRender[F[_] : AsValueObservable]: AsVDomModifier[F[VDomModifier]] = (valueStream: F[VDomModifier]) =>
-    ModifierStreamReceiver(ValueObservable(valueStream))
+    ModifierStreamReceiver(ValueObservable.from(valueStream))
+
+  implicit def ValueObservableRenderAs[T : AsVDomModifier]: AsVDomModifier[ValueObservable[T]] = (valueStream: ValueObservable[T]) =>
+    ModifierStreamReceiver(valueStream.map(VDomModifier(_)))
 
   implicit def valueObservableRenderAs[T : AsVDomModifier, F[_] : AsValueObservable]: AsVDomModifier[F[T]] = (valueStream: F[T]) =>
-    ModifierStreamReceiver(ValueObservable(valueStream).map(VDomModifier(_)))
+    ModifierStreamReceiver(ValueObservable.from(valueStream).map(VDomModifier(_)))
 
   implicit def childCommandObservableRender[F[_] : AsValueObservable]: AsVDomModifier[F[ChildCommand]] = (valueStream: F[ChildCommand]) =>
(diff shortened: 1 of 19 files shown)


src/main/scala/outwatch/Handler.scala
```diff
@@ -14,26 +16,30 @@ object Handler {
...
 
 object ProHandler {
   def create[I,O](f: I => O): IO[ProHandler[I,O]] = for {
     handler <- Handler.create[I]
   } yield handler.mapObservable[O](f)
 
-  def apply[I,O](observer:Observer[I], observable: Observable[O]):ProHandler[I,O] = new Observable[O] with Observer[I] {
+  def apply[I,O,F[_]: AsValueObservable](observer:Observer[I], observable: F[O]):ProHandler[I,O] = apply(observer, ValueObservable(observable))
+  def apply[I,O](observer:Observer[I], valueObservable: ValueObservable[O]):ProHandler[I,O] = new ValueObservable[O] with Observer[I] {
     override def onNext(elem: I): Future[Ack] = observer.onNext(elem)
```
fdietze: We could inline these redirection methods

cornerman:
I think, this is a version which should work well, now. Though, there are some implications for the users:

You should generally use the `ValueObservable`. For example, you might have some function for a component: `def component(source: Observable[T]): VNode`. Whenever you now pass in a handler (`Handler.create[T].map(component(_))`), then the handler will be happily converted to an `Observable` through an implicit from ValueObservable to Observable. But this conversion might not be desired, you actually want to depend on a ValueObservable, so you get an efficient first render for your component. So, you should write your component like this: `def component(source: ValueObservable[T]): VNode`.

We can discuss whether, we should remove the implicit, so that you have to explicitly call `toObservable` on a ValueObservable. The advantage of the implicit is that you can use any function defined on an Observable on the ValueObservable (we just provide a subset of common operations). Also you can pass a handler down to anything where an Observable is required. Well, the disadvantage now is that you might mistakenly convert a ValueObservable to an Observable, which is not incorrect but a bit less efficient.

cornerman:
I have now removed the implicit from `ValueObservable` to `Observable` - and found a bug because of it :)

The `ValueObservable` interface now has some more convenience methods. you can call `.observable` on it to get the dual `Observable` that you can use for more complex stuff or interop with other code.

cornerman:
This PR works and gives a performance boost for most applications, but we should really think about it. For the user, it introduces another type `ValueObservable`, which is to be used instead of `Observable`. And we need to maintain the mirroring API to reflect the one of Observable. I am not sure, it is worth. Probably better to just use another streaming library for state mangement.

cornerman:
I still like this PR and it measurably improves performance in my benchmarks by about 20%, but I do not like mirroring the `Observable` API in our `ValueObservable` trait. It is tested but it is potentially another source of bugs and we do really not want to maintain the API here. Maybe it shows that monix is not the right tool for state management and underlines the importance of #96. We will want to improve on our typeclasses `AsValueObservable` and `AsObserver` as well as make a proper documentation on this topic. Then we can discuss what frp library we should use for the core.

Therefore, I will close this PR, because I do not think we should merge it anymore.

Issue #266, opened 2018-11-27
Cleanup OutWatchDomSpec


cornerman:
This file is huge and mixes all kind of test-cases. It could be split into multiple files that cover different areas of the API.

I feel like I just keep on appending to this file :)

marklister:
I had a go at this... It's not as easy as it sounds... thus far i have something like this:

```dom
   AsVDomModifierSpec
   DomPatchSpec
   HtmlDslSpec
   LocalStorageSpec
   NodeKeySpec
   SeparatedModifiersSpec
   SnabdomOpsSpec
   VNodeSpec```

Unfortunately most tests (I've moved around half, but i see myself restarting from scratch) conform to a pattern that places them in DomPatchSpec or HtmlDslSpec.  @cornerman  thoughts?

cornerman:
@marklister Looks good to me! Don't worry about a lot of tests going into the same file in the end. It is an improvement if we can separate different topics like `LocalStorageSpec`, `AsVDomModifierSpec`, `SnabbdomOpsSpec`, `SeparatedModifierSpec`, `NodeKeySpec` and `DomPatchSpec`/`HtmlDslSpec`.

Sometimes it is probably difficult to tell where a test could belong. Maybe for now, we just split out the ones that are cleary about something else than dom patching. Afterwards, we can reiterate and check if we see another pattern. But in the end, the dom patching tests will make up a lot of the tests.

PR #270, opened 2018-12-06
Multiple changes to Store

Busti:
This PR makes the following changes to `Store` (in order of significance)

* Makes Store emit an (action -> state) tuple.
  * This tuple allows more granular control about *when* a view updates when the stores state changes.  
    Methods like `collect` can now be used to, not only capture certain state changes, but they can also react to the `Action` which caused that state.  
    This allows us to ignore all other Actions which are irrelevant to that specific view.  
    For example, if we were to move a draggable element using a mouse cursor, we could react just to events changing the position of the element, while discarding all other state updates.
    ```scala
    svg(
      rect(width := 10, height := 10,
        x <-- store.collect { case (MoveAction, state) => state.xPos },
        y <-- store.collect { case (MoveAction, state) => state.yPos }
      )
    )
    ```
  * This tuple can also be used to have other parts of the application react to an `Action`. This can be useful if we, for example want to collect numerous actions (and their resulting states) into a list, so that we can later undo them, as as is common in many applications, by, for example, hitting `ctrl + z`.
  * However this addition has some implications which have further consequences.
* Introduces a common Action base-trait;
  * Since we now emit an `(action -> state)` tuple in `Store`, we have 2 edge cases which need to be solved.
    * This PR Introduces a common starting Action: Noop;
      * When created, `Store` emits it's initial State once, to every subscriber. Since that state was not determined by an Action, a placeholder action is needed.
        It, `Noop` is inherited from the common `Action`.
    * This PR Introduces a common Error Action which carries error data.  
      * The `Error()` Action is used, when the reduction of an Action fails. It contains the errors data object.
* Changes the Store type parameter naming scheme, to use Model instead of State.
  * I propose the type parameters of store to be named `[A, M]`, meaning `[Action, Model]` (ordered by `In, Out]`, consistently, as the ordering is also used in `ProHandler`)
    I feel like this describes the usage of these a little better, since we do not actually have a state at compile time.  
    A runtime value should be named `val state: Model`
* Removes startWith by using the scan0 operator;

---

I am afraid, that I cannot really split this into multiple commits, without a lot of work being involved, since my original branch is unable to merge due to it being forked from the `new-io`  PR while it was still in progress.

### DIFF
--- a/outwatch/src/main/scala/outwatch/util/Store.scala
+++ b/outwatch/src/main/scala/outwatch/util/Store.scala
@@ -28,31 +28,34 @@ object Store {
   }
 
   def create[A, M](
+    initialAction: A,
     initialState: M,
     reducer: Reducer[A, M]
-  )(implicit s: Scheduler): IO[ProHandler[A, M]] = IO {
+  )(implicit s: Scheduler): IO[ProHandler[A, (A, M)]] = IO {
+    val subject = PublishSubject[A]
 
-      val subject = PublishSubject[A]
-
-      val fold: (M, A) => M = (state, action) => Try { // guard against reducer throwing an exception
+    val fold: ((A, M), A) => (A, M) = {
+      case ((_, state), action) => Try { // guard against reducer throwing an exception
         val (newState, effects) = reducer.reducer(state, action)
 
         effects.subscribe(
-          e => subject.feed(e :: Nil),
-          e => dom.console.error(e.getMessage) // just log the error, don't push it into the subject's observable, because it would stop the scan "loop"
+          next => subject.feed(next :: Nil),
+          // just log the error, don't push it into the subject's observable, because it would stop the scan "loop"
+          error => dom.console.error(error.getMessage)
         )
-        newState
+
+        action -> newState
       }.recover { case NonFatal(e) =>
         dom.console.error(e.getMessage)
-        state
+        action -> state
       }.get
+    }
 
-      subject.transformObservable(source =>
-        source
-          .scan(initialState)(fold)
-          .share
-          .startWith(Seq(initialState))
-      )
+    subject.transformObservable(source =>
+      source
+        .scan0[(A, M)](initialAction -> initialState)(fold)
+        .replay(1).refCount
+    )
   }
 
   private val storeRef = STRef.empty
@@ -60,13 +63,18 @@ object Store {
   def get[A, M]: IO[ProHandler[A, M]] = storeRef.asInstanceOf[STRef[ProHandler[A, M]]].getOrThrow(NoStoreException)
 
   def renderWithStore[A, M](
-    initialState: M, reducer: Reducer[A, M], selector: String, root: VNode
+    initialAction: A,
+    initialState: M,
+    reducer: Reducer[A, M],
+    selector: String,
+    root: VNode
   )(implicit s: Scheduler): IO[Unit] = for {
-    store <- Store.create(initialState, reducer)
-    _ <- storeRef.asInstanceOf[STRef[ProHandler[A, M]]].put(store)
+    store <- Store.create[A, M](initialAction, initialState, reducer)
+    _ <- storeRef.asInstanceOf[STRef[ProHandler[A, M]]].put(store.mapProHandler[A, M](in => in)(out => out._2))
     _ <- OutWatch.renderInto(selector, root)
   } yield ()
 
   private object NoStoreException
     extends Exception("Application was rendered without specifying a Store, please use Outwatch.renderWithStore instead")
+
 }
--- a/outwatch/src/main/scala/outwatch/util/Store.scala
+++ b/outwatch/src/main/scala/outwatch/util/Store.scala
@@ -23,8 +23,8 @@ object Store {
     implicit def justState[M, A](f: (M, A) => M): Reducer[M, A] = Reducer { (s: M, a: A) => (f(s, a), Observable.empty) }
 
     implicit def stateAndOptionIO[M, A](f: (M, A) => (M, Option[IO[A]])): Reducer[M, A] =  Reducer { (s: M, a: A) =>
-      val (mewState, effect) = f(s, a)
-      (mewState, effect.fold[Observable[A]](Observable.empty)(Observable.fromIO))
+      val (newState, effect) = f(s, a)
+      (newState, effect.fold[Observable[A]](Observable.empty)(Observable.fromIO))
     }
   }
 
--- a/outwatch/src/main/scala/outwatch/util/Store.scala
+++ b/outwatch/src/main/scala/outwatch/util/Store.scala
@@ -30,7 +30,8 @@ object Store {
   def create[A, M](
     initialAction: A,
     initialState: M,
-    reducer: Reducer[A, M]
+    reducer: Reducer[A, M],
+    recoverError: PartialFunction[Throwable, M => M] = PartialFunction.empty
   )(implicit s: Scheduler): IO[ProHandler[A, (A, M)]] = IO {
     val subject = PublishSubject[A]
 
@@ -45,10 +46,10 @@ object Store {
         )
 
         action -> newState
-      }.recover { case NonFatal(e) =>
+      }.recover(recoverError.andThen(f => action -> f(state)) orElse { case NonFatal(e) =>
         dom.console.error(e.getMessage)
         action -> state
-      }.get
+      }).get
     }
 
     subject.transformObservable(source =>
--- a/outwatch/src/main/scala/outwatch/util/Store.scala
+++ b/outwatch/src/main/scala/outwatch/util/Store.scala
@@ -15,19 +15,52 @@ import scala.util.control.NonFatal
 
 object Store {
 
+  /**
+   * A Function that applies an Action onto the Stores current state.
+   * @param reducer The reducing function
+   * @tparam A The Action Type
+   * @tparam M The Model Type
+   */
   case class Reducer[A, M](reducer: (M, A) => (M, Observable[A]))
 
   object Reducer {
+    /**
+     * Creates a Reducer which yields a new State, as-well as an Observable of Effects
+     * Effects are Actions which will be executed after the Action that caused them to occur.
+     * This is accomplished by subscribing to the Effects Observable within the stores scan loop.
+     *
+     * CAUTION: There is currently a bug which causes the Effect-States to emit,
+     * before the State of the action that caused the effects is emitted.
+     * However, this only effects immediate emissions of the Effects Observable, delayed emissions should be fine.
+     * @param f The Reducing Function returning the (Model, Effects) tuple.
+     */
     implicit def stateAndEffects[A, M](f: (M, A) => (M, Observable[A])): Reducer[A, M] = Reducer(f)
 
+    /**
+     * Creates a reducer which just transforms the state, without additional effects.
+     */
     implicit def justState[A, M](f: (M, A) => M): Reducer[A, M] = Reducer { (s: M, a: A) => (f(s, a), Observable.empty) }
 
+    /**
+     * Creates a Reducer with an optional IO effect.
+     */
     implicit def stateAndOptionIO[A, M](f: (M, A) => (M, Option[IO[A]])): Reducer[A, M] = Reducer { (s: M, a: A) =>
       val (newState, effect) = f(s, a)
       (newState, effect.fold[Observable[A]](Observable.empty)(Observable.fromIO))
     }
   }
 
+  /**
+   * Creates an IO[Store]
+   * A Store capules a scan operation on an Observable in an opinionated manner.
+   * An internal state is tansformed by a series of actions.
+   * The state will be the same for every subscriber.
+   * @param initialAction The Stores initial action. Useful for re-creating a store from memory.
+   * @param initialState The stores initial state. Similar to the initial accumulator on a fold / scan.
+   * @param reducer The Reducing funcion. Creates a new State from the previous state and an Action.
+   * @param recoverError An optional PartialFunctio for recovering the State when an Exception occurs in the Recuer.
+   * @return An Observable emitting a tuple of the current state and the action that caused that state.
+   */
   def create[A, M](
     initialAction: A,
     initialState: M,
@@ -65,10 +98,22 @@ object Store {
     out
   }
 
+  /**
+   * A global reference to a Store.
+   * Commonly used to implement the Redux Pattern.
+   */
   private val storeRef = STRef.empty
 
+  /**
+   * Get's a globally unique store.
+   * Commonly used to implement the Redux Pattern.
+   */
   def get[A, M]: IO[ProHandler[A, M]] = storeRef.asInstanceOf[STRef[ProHandler[A, M]]].getOrThrow(NoStoreException)
 
+  /**
+   * Renders an Application with a globally unique Store.
+   * Commonly used to implement the Redux Pattern.
+   */
   def renderWithStore[A, M](
     initialAction: A,
     initialState: M,
--- a/outwatch/src/main/scala/outwatch/util/Store.scala
+++ b/outwatch/src/main/scala/outwatch/util/Store.scala
@@ -1,6 +1,7 @@
 package outwatch.util
 
 import cats.effect.IO
+import monix.eval.Task
 import monix.execution.Scheduler
 import monix.reactive.Observable
 import monix.reactive.subjects.PublishSubject
@@ -52,11 +53,16 @@ object Store {
       }).get
     }
 
-    subject.transformObservable(source =>
+    val out = subject.transformObservable(source =>
       source
         .scan0[(A, M)](initialAction -> initialState)(fold)
         .replay(1).refCount
     )
+
+    val sub = out.subscribe()
+    out.doOnSubscribeF(Task(sub.cancel))
+
+    out
   }
 
   private val storeRef = STRef.empty
--- a/outwatch/src/main/scala/outwatch/util/Store.scala
+++ b/outwatch/src/main/scala/outwatch/util/Store.scala
@@ -15,15 +15,14 @@ import scala.util.control.NonFatal
 
 object Store {
 
-  case class Reducer[S, A](reducer: (S, A) => (S, Observable[A]))
+  case class Reducer[M, A](reducer: (M, A) => (M, Observable[A]))
 
   object Reducer {
+    implicit def stateAndEffects[M, A](f: (M, A) => (M, Observable[A])): Reducer[M, A] = Reducer(f)
 
-    implicit def stateAndEffects[S, A](f: (S, A) => (S, Observable[A])): Reducer[S, A] = Reducer(f)
+    implicit def justState[M, A](f: (M, A) => M): Reducer[M, A] = Reducer { (s: M, a: A) => (f(s, a), Observable.empty) }
 
-    implicit def justState[S, A](f: (S, A) => S): Reducer[S, A] = Reducer { (s: S, a: A) => (f(s, a), Observable.empty) }
-
-    implicit def stateAndOptionIO[S, A](f: (S, A) => (S, Option[IO[A]])): Reducer[S, A] =  Reducer { (s: S, a: A) =>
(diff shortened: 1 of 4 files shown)


outwatch/src/main/scala/outwatch/util/Store.scala
```diff
@@ -14,58 +14,63 @@ import scala.util.control.NonFatal
 
 
 object Store {
+  sealed trait BaseAction
+  private final case object Noop extends BaseAction
+  private final case class Error(context: Action, error: Throwable) extends BaseAction
+  trait Action extends BaseAction
 
-  case class Reducer[S, A](reducer: (S, A) => (S, Observable[A]))
+  case class Reducer[A, M](reducer: (M, A) => (M, Observable[A]))
```
cornerman: Should the order of argument here be consistent with the type argument order? Like having `[A,M]` and `(M, A)`.

Thinking about this order change might break existing user code. Does it still make sense to do it?

cornerman:
@Busti Thank you for you contribution :)

In general this looks good to me. But before merging I would like to look out for alternatives of having a common base trait for actions.

What do you think about a typeclass like the following?
```scala
trait Action[T] {
  def noop: T
  def error(err: Throwable): T
}
```

We could even have a simple default implementation with an `Option` and just throw on error :)

cornerman:
Wait. The user provides the `initialState` in the `renderWithStore` method, so how about we just let the user provide the tuple of action and state there? We can have another method with just the state which create a `Store[Option[A], M]`. Then we might not even need an action type at all.

outwatch/src/main/scala/outwatch/util/Store.scala
```diff
@@ -14,58 +14,63 @@ import scala.util.control.NonFatal
...
+      val subject = PublishSubject[A]
+
+      val fold: ((BaseAction, M), A) => (BaseAction, M) = {
+        case ((_, state), action) => Try { // guard against reducer throwing an exception
+          val (newState, effects) = reducer.reducer(state, action)
+
+          effects.subscribe(
+            next => subject.feed(next :: Nil),
+            // just log the error, don't push it into the subject's observable, because it would stop the scan "loop"
+            error => dom.console.error(error.getMessage)
```
ahjohannessen: Perhaps more flexible to use an `ErrorSink` instead `dom.console.error`, or at least have a way to turn logging off.

cornerman: Good idea! But it does not necessarily need to be fixed in this PR.

ahjohannessen: Just a thought :)

outwatch/src/main/scala/outwatch/util/Store.scala
```diff
@@ -14,58 +14,63 @@ import scala.util.control.NonFatal
 
 
 object Store {
+  sealed trait BaseAction
```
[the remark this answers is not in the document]
Busti: I have to admit that I do not like it all that much either, but, in order to represent all 3 possible action types we would have to use something like...
```scala
  type Action = Option[Try[A]]
```

Since the previous action could also be an error. This would make the reducer look rather verbose:  
```scala
val reducer = {
  case (state, Some(Success(IncrementBy(number)))) => state += number
}
```

Busti:
> @Busti Thank you for you contribution :)
> 
> In general this looks good to me. But before merging I would like to look out for alternatives of having a common base trait for actions.
> 
> What do you think about a typeclass like the following?
> 
> ```scala
> trait Action[T] {
>   def noop: T
>   def error(err: Throwable): T
> }
> ```
> 
> We could even have a simple default implementation with an `Option` and just throw on error :)

I must admit that I really like not having to wrap each pattern match `case` in some wrapper trait, but there is probably a nicer way to accomplish that than using a common base trait.

outwatch/src/main/scala/outwatch/util/Store.scala
Busti: I think thr order here should be the same as the order commonly used in `fold`.

Busti:
@cornerman May I ask, why do you suggest a typeclass?  
I do not see any case where this would need any implicit type-depending behavior.  
We could probably just model it after scalas `Option` or `Either`, but I do not like creating new wrapper classes like that. There is too much stuff we'd have to implement.  

I suppose we could just go for `Option[Try[T]]` and then implement some useful implicits that make it nicer to use.

Busti:
I like the Idea of providing an action when passing the `initialState` to the store. I have a lot of scenarios where store contents are provided by a server, because they were created earlier on the client.

outwatch/src/main/scala/outwatch/util/Store.scala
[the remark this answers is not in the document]
Busti: I like having the `Action` as context to the `Throwable`. It would be rather useful when, for example, displaying a toast message to the user.

fdietze:
Offtopic:
@Busti for future work / curiosity. Like two months ago I experimented with an incremental Store concept to calculate efficient updates to the dom. The key concept is similar to your idea: The store store outputs state and action, and when you map, you can map over both to transform state and action at the same time. This allows to calculate efficient updates to the state and later to the dom. I gave it up, because it turned out that incremental components are hard to write. But I want to show it to you anyways. Maybe you have an idea.
(I dont't remember if this code is in a usable state, but you should get the idea.)

```scala
import monix.execution.{Ack, Cancelable, Scheduler}
import monix.reactive.observers.Subscriber
import monix.reactive.{Observable, Observer}
import monix.reactive.subjects.PublishSubject
import rx._

import scala.concurrent.Future

trait IncStore[State, InAction, OutAction] { self =>
  def initialState: State
  def inAction: Observer[InAction]
  def output: Observable[(State, OutAction)]
  val state: Observable[State] = output.map(_._1)
  val outAction: Observable[OutAction] = output.map(_._2)

  def mapIncremental[NextState, NextOutAction](mapInputState:State => NextState, onAction:(NextState,OutAction) => (NextState, NextOutAction)): IncStore[NextState, InAction, NextOutAction] = new IncStore[NextState, InAction, NextOutAction] {
    override def initialState: NextState = mapInputState(self.initialState)
    override def inAction: Observer[InAction] = self.inAction
    override def output: Observable[(NextState, NextOutAction)] = self.output.map { case (state, action) => onAction(mapInputState(state), action)}
  }

  def map[NextState](mapInputState:State => NextState, onAction:(NextState,OutAction) => NextState): IncStore[NextState, InAction, OutAction] = new IncStore[NextState, InAction, OutAction] {
    override def initialState: NextState = mapInputState(self.initialState)
    override def inAction: Observer[InAction] = self.inAction
    override def output: Observable[(NextState, OutAction)] = self.output.map { case (state, action) => (onAction(mapInputState(state), action), action) }
  }

  def mapState(mapInputState: State => State): IncStore[State, InAction, OutAction] = new IncStore[State, InAction, OutAction] {
    override def initialState: State = mapInputState(self.initialState)
    override def inAction: Observer[InAction] = self.inAction
    override def output: Observable[(State, OutAction)] = self.output.map { case (state, action) => (mapInputState(state), action) }
  }
}

case class ConcreteIncStore[State, InAction, OutAction](initialState: State, inAction: Observer[InAction], output: Observable[(State, OutAction)]) extends IncStore[State, InAction, OutAction]

object IncStore {
  type Store[State, Action] = IncStore[State, Action, Action]

  def apply[State, Action](init:State, f:(State,Action) => State): Store[State, Action] = {
    val subject = PublishSubject[Action]()
    var lastState = init
    val observable = subject.map { inAction =>
      lastState = f(lastState, inAction)
      (lastState, inAction)
    }
    new ConcreteIncStore[State, Action, Action](init, subject, observable)
  }
  def incremental[State, InAction, OutAction](init: State, f:(State,InAction) => (State, OutAction)): IncStore[State, InAction, OutAction] = {
    val subject = PublishSubject[InAction]()
    var lastState = init
    val observable = subject.map { inAction =>
      val result = f(lastState, inAction)
      lastState = result._1
      result
    }
    new ConcreteIncStore[State, InAction, OutAction](init, subject, observable)
  }

object StoreExample {
  def sorted(implicit ctx: Ctx.Owner): VNode = {
    sealed trait SetAction
    object SetAction {
      case class Add(value: Int) extends SetAction
      case class Remove(value: Int) extends SetAction
    }
    sealed trait SeqAction
    object SeqAction {
      case class Add(value: Int, index: Int) extends SeqAction
      case class Remove(index: Int) extends SeqAction
    }
    val intList = Store[List[Int], SetAction](Nil, {
      case (list, action@SetAction.Add(elem))    => elem :: list
      case (list, action@SetAction.Remove(elem)) => list diff List(elem)
    })

    val (sortedIntSeq, _) = intList.mapIncremental[Vector[Int], SeqAction](_.sorted.toVector, {
      case (currentVector, SetAction.Add(nextElem)) =>
        val (smaller, bigger) = currentVector.partition(_ < nextElem) // can be made O(log n)
      val position = smaller.size
        val newState = (smaller :+ nextElem) ++ bigger
        (newState, SeqAction.Add(nextElem, position))
      case (currentVector, SetAction.Remove(elem))  =>
        val position = currentVector.indexOf(elem) // can be made O(log n)
      val newState = currentVector.take(position) ++ currentVector.drop(1 + position)
        (newState, SeqAction.Remove(position))
    })

    val sortedDivs: Observable[ChildCommand] = sortedIntSeq.outAction.map {
      case SeqAction.Add(elem, index) =>
        val node = div(
          display.flex,
          elem,
          div("x", paddingLeft := "10px", cursor.pointer, onClick.mapTo(SetAction.Remove(elem)).handleWith(intList.update _))
        )
        ChildCommand.Insert(index, node)
      case SeqAction.Remove(index)    =>
        ChildCommand.Remove(index)
    }

    div(
      button("Add random number", onClick.mapTo(SetAction.Add(scala.util.Random.nextInt(100))).handleWith(intList.update _)),
      div("intList: ", intList.state.map(_.toString), ", last action: ", intList.outAction.map(_.toString)),
      div("sorted: ", sortedIntSeq.state.map(_.toString), ", last action: ", sortedIntSeq.outAction.map(_.toString)),
      div(sortedDivs.startWith(Seq(ChildCommand.ReplaceAll(js.Array(div("initial element")))))),
    )
  }
}
```

Busti:
I will implement it using a `Try` for now.  
@cornerman I have to agree with you on your previous point. A `Noop` is unnecessary when we provide an initial action.

cornerman:
> I will implement it using a Try for now.

Great. Maybe we should prefer `Either[Throwable, T] over `Try[T]`? It just behaves more intuitive to me, because it does not swallow exceptions like `Try` when composing it.

Example:
```scala
scala> Try(1).map(_ => throw new Exception)
res1: scala.util.Try[Nothing] = Failure(java.lang.Exception)
```

```scala
scala> Right[Throwable, Int](1).map(_ => throw new Exception)
java.lang.Exception
  at .$anonfun$res3$1(<console>:15)
  at .$anonfun$res3$1$adapted(<console>:15)
  at scala.util.Either.map(Either.scala:349)
  ... 28 elided
```

outwatch/src/main/scala/outwatch/util/Store.scala
marklister: @Busti I think this is what @cornerman is trying to say he prefers:

`case class Reducer[M, A](reducer: (M, A) => (M, Observable[A]))`  

I also find this order preferable.  Agree that the fold pattern is ubiquitous.

Busti:
@cornerman I do not like it all that much either, since `Try` also swallows the exception type. I would still like to pass the failed action as context, since it could be useful when displaying errors, but that depends on how likely it is that a user causes an error.

cornerman:
> I do not like it all that much either, since `Try` also swallows the exception type. I would still like to pass the failed action as context, since it could be useful when displaying errors, but that depends on how likely it is that a user causes an error.

On the other hand, one should not put too much logic into exceptions. Users should rather model these reducer errors in their state, which could include a potential error together with the previous state.

What I am trying to say: If  we want to know the action that lead to an error, maybe it makes sense to put the error in the state and even make this optional with a `PartialFunction`. You can handle all errors and get to know the action that lead to it and can even reflect the error in your state if you want to:

```scala
  def create[A, M](
    initialAction: A,
    initialState: M,
    reducer: Reducer[A, M],
    recoverError: PartialFunction[Throwable, M => M] = PartialFunction.empty
  )(implicit s: Scheduler): IO[ProHandler[A, (A, M)]] = IO {
   ...
   }.recover(recoverError.andThen(f => f(state)) orElse { case NonFatal(e) =>
     dom.console.error(e.getMessage)
     Failure(e) -> state
   }).get
}
```

Busti:
@cornerman I like the general direction of the Idea. I had some second thoughts about it that I would like to share, however I have come to the conclusion that this is the best solution.  
The

cornerman:
> Also, shouldn't the default case for the PartialFunction be { case _ => s => s }, since PartialFunction.empty throws a MatchError every time it is called.

Both would work. We just add a console error if you do not handle the error yourself. The `PartialFunction.empty` does not throw a match error, because we never do an `apply()` on it. We use `andThen` which returns another PartialFunction and then do `orElse` which returns another PartialFunction. Therefore the `recover` of the monix observable will just see a PartialFunction and if `recoverError` is not defined, it will fallback to the orElse branch (`dom.console.error` and returning `action -> state`).

Busti:
I implemented the previously discussed changes. Are there any other matters that need to be discussed?

cornerman:
Great, it looks good to me.

I think, we can improve the store even more, but this does not need to be done in this PR :)

fdietze:
@Busti Can you please squash and rebase? Then we can merge. Thank you for your contribution!

Busti:
I am afraid that there still is a bug in the current impl. I have been doing some testing with this in another project. I will write a unit-test for the bug I found, but unfortunately I have been unable to fix it yet.  

The problem is as follows:
When the stores emissions are used in multiple locations, only the first usage ever receives the initial value.
```scala
for {
  store <- Store.create[Int, String](0, "foo", {
      case (_, 0) => "foo"
      case (_, _) => "bar"
    }: (String, Int) => String)

  // Unfortunately a ProHandler cannot be used as an observable in tags.
  state = store.mapObservable(_._2).asinstanceOf[Observable[String]]

  OutWatch.renderReplace("#app", ul(
    li(test),
    li(test)
  ))
} yield ExitCode.Success
```

This should yield:
```html
<ul>
    <li>foo</li>
    <li>foo</li>
</ul>
```

This does yield:
```html
<ul>
    <li>foo</li>
    <li></li>
</ul>
```

cornerman:
> // Unfortunately a ProHandler cannot be used as an observable in tags. `state = store.mapObservable(_._2).asinstanceOf[Observable[String]]`

A ProHandler is just an `Observable with Observer`, so it should work like that (see https://github.com/OutWatch/outwatch/blob/master/outwatch/src/main/scala/outwatch/MonixOps.scala#L11). The `asInstanceOf` should therefore not be needed for the scala compiler. Is it intellij complaining?

You can write: ```scala
store.map(_._2) // this just creates an Observable. mapObservable creates a new ProHandler where only the Observable part is mapped: ProHandler[(A,B), B]
```

**Comment by cornerman** [comment]
> Changing the multicast method to behavior seems to fix the issue. But maybe there is a cleaner solution.

Interesting! Yeah that is a good point with multicasting observables. We need to assure that you get your initial action on each subscription. But I am not sure whether `BehaviorSubject` is the right approach here, because we would only get the latest value as first value (depending at what point in time you are subscribing to the observable).  Furthermore, here `behavior(...)` returns a `ConnectableObservable` to which you need to `connect()` to make the observable work properly.

We need to be aware that we are subscribing in each call to the `fold` method. So if we do not use `share` here, we would subscribe twice on each new value because the whole observable is computed for each subscriber.

How about always starting with the initialValue and doing a `share` before (but of course you would not get the the latest state if you subscribe late):

```scala
  .scan(...)
  .share
  .prepend(initalAction -> initialState)
```

Busti:
You are right, It is actually just IntelliJ that is complaining, with:  
```
Type mismatch, expected: VDomModifier, actual: dom.ProHandler[A, M]
```

I should really learn to compile, before taking IntelliJ highlights for granted.

Busti:
Prepending the initial event would not work, when there were "very late subscribers", i.e. subscriptions that are only made after the first (not initial) action has been processed by the Store.  
If such a subscription would occur, it would not display the current state, until another Action would have been processed.

Busti:
I have updated `Store` and the test to reflect this scenario.  
I found that `replay(1).refCount` does work as intended, when used as a multicast method.  

I have introduced a new variable in the counter model, which counts, how often a number has been added or subtracted.  
I would like to create some more granular tests for store. Which spec file should I create these in?

cornerman:
> I would like to create some more granular tests for store. Which spec file should I create these in?

Cool, maybe a dedicated `StoreSpec` file would make sense.

Busti:
I ended up calling it `UtilSpec`, since I have an upcoming PR which should also be tested in there.  
Given the general style of tests being in very few files, I do not think that `Store` justifies its own test file.  

I ended up using `WordSpec` since it removes a lot of redundancy from the tests, but I can change it to `FlatSpec` if you prefer that.

cornerman:
Thanks for the update!

> Given the general style of tests being in very few files...

I do not like this style and would therefore appreciate having a dedicated `StoreSpec`. I hope, we will generally proceed in having more specific test files in the future (see https://github.com/OutWatch/outwatch/issues/266).

> I also ended up using WordSpec since it removes a lot of redundancy from the tests, but I can change it to FlatSpec if you prefer that.

I would like to keep the spec style consistent accross files, so it would be great if you could rewrite it to use `FlatSpec`. Eventhough, I am not a fan of this specific style, it is how the other files were written.

Busti:
Sure, I have changed the filename and the test-style.  

There were two things that came up when I made the changes, one of which I was able to solve.  
1. The store does not handle actions until the first subscription is made.
  I solved this by making a pseudo subscription to the store which is canceled right after he first actual subscription is made.
2. When the reducer returns an `Observable.from(Seq(Action, Action))` those actions are computed and emitted before the action that caused them is emitted, however, the action that caused them is emitted afterwards, which makes the emissions really inconsistent.
  If we, for example, had a counter application, which emits two `Plus` actions for every `Minus` action, we would see something like this: `0, 1, 2, -1` instead of the expected `0, -1, 0, 1`  
  I have written a test for this scenario, which I am going to commit next, but I was not able to find a solution just yet.

Busti:
One solution I have thought of would be to return the `effect` observable in the `fold` function and then move the subscription to the `effects` outside of the scan loop, after it completes.  
```scala
val fold: ((A, M), A) => ((A, M), Observable[A])

subject.transformObservable( source =>
  source
    .scan0[(A, M)](initialAction -> initialState)(fold)
    .map { case (state, effects) =>
      effects.subscribe(
        next => subject.feed(next :: Nil),
        error => dom.console.error(error.getMessage)
      )
      state
    }
    .replay(1).refCount
```

Busti:
This is what I had in mind, but for some reason it behaves even more unpredictable.  
The test from before now outputs `0, 0, -1, 1` instead of the expected `0, -1, 0, 1`

cornerman:
> I would appreciate some input on this.

Interesting problem! I will check it out in more detail on the weekend and let you know what I can find :)

Busti:
Here is a link to the fiddle and its code for reference.
https://scalafiddle.io/sf/7tgDepe/1

```scala
val in = PublishSubject[Int]

val out = PublishSubject[Int]

in.scan((0, Observable.empty[Int])){
  case ((s, _), 0) => (1, Observable.fromIterable(Seq(2, 3)))
  case ((s, _), n) => (s + n, Observable.empty)
}.foreach { case (s, effects) =>
  out.onNext(s)
  effects.foreach(in.onNext)
}

out.foreach(println)

in.onNext(0)
```

Other than added complexity, I cannot see any difference between the fiddle and the actual impl.

Busti:
Since there is no real end to this in sight, I propose that we leave this bug out of this PR.  
It already is way too big as it is. I will squash this into 2 - 3 Commits that make sense and the bug can be resolved in another PR.

Busti:
Please have a look at the new commit structure.  
The intermediate commits have been removed and every change has it's own commit.  

I have also added some documentation.

cornerman:
@Busti Great! The PR looks good now. Additionally thank you for the nice documentation.

Can you open an issue for the bug? Just to have some info on what you have found out.

PR #275, opened 2018-12-30
Make EmitterBuilder.apply call-by-name

fdietze:
I just found a lot of bugs in my code, realizing that `EmitterBuilder.apply` is not call-by-name.

In my opinion it would be more intuitive if it was call-by-name, so one can write:
```scala
onClick(util.Random.nextInt) --> ...
```

### DIFF
--- a/outwatch/src/main/scala/outwatch/dom/helpers/EmitterBuilder.scala
+++ b/outwatch/src/main/scala/outwatch/dom/helpers/EmitterBuilder.scala
@@ -22,7 +22,7 @@ trait EmitterBuilder[+O, +R] { self =>
 
   @inline def foreach(action: O => Unit): R = -->(Sink.fromFunction(action))
   @inline def foreach(action: => Unit): R = foreach(_ => action)
-  @inline def apply[T](value: T): EmitterBuilder[T, R] = map(_ => value)
+  @inline def apply[T](value: => T): EmitterBuilder[T, R] = mapTo(value)
   @inline def mapTo[T](value: => T): EmitterBuilder[T, R] = map(_ => value)
   @inline def apply[T](latest: Observable[T]): EmitterBuilder[T, R] = transform(_.withLatestFrom(latest)((_, u) => u))
   @inline def debounce(timeout: FiniteDuration): EmitterBuilder[O, R] = transform(_.debounce(timeout))


cornerman:
This is why we have mapTo. Imho apply should not be call by name, because it would be executed ever time.

LukaJCB:
I'm also :-1: on this, `Random.nextInt` is a side effect, if you want to have lazy evaluation because of an undeterministic method call, then that is exactly why we use mechanisms like `IO` in the first place!

fdietze:
Ok, I have to admit that the random number was a bad example.

The use-case we had in our project was with scala.rx, where we wrote `onClick(someRx.now) --> ... `. Today I grepped through our code and found 7 Usages like this. Some were written by @cornerman :wink:. All were bugs.

A new user to whom I showed outwatch was also very confused by this behavior.

People write functions in there, that calculate something and write the value into a sink. They expect the function to be executed on every click. Then they don't find the mistake.

What I want to communicate, is that mistakes are being made with the current API. Simply saying that you should not do mistakes is dangerous in my opinion. We should design the API, so that the risk of making a mistakes is minimized.

LukaJCB:
> What I want to communicate, is that mistakes are being made with the current API. Simply saying that you should not do mistakes is dangerous in my opinion. We should design the API, so that the risk of making mistakes is minimized.

I fully agree with the sentiment.
However,
> People write functions in there, that calculate something and write the value into a sink. They expect the function to be executed on every click. Then they don't find the mistake in their code.

This will lead down a rabbit hole that's hard to get out of. The only reason this is making any difference at all is because the function is side effecting. This never happens at all for pure functions and taking a by-name argument is a band-aid solution IMO. How do we design our API so that the risk of making mistakes is minimized?
This is a hard question to answer, but I don't think giving people the tools to abuse side-effects in this regard is a good answer. I think side-effects are much more prone to lead to bugs and should therefore be made as explicitly as possible. If we have to consider that any function any user will ever give us can be side-effecting, then we have to design our API in a whole different way.
IMO OutWatch should be opinionated in this regard, if you want side-effects make them explicit (like with `IO`, `foreach` or `SideEffects`. 

Of course, I've been mostly absent from maintenance of this project and I do apologize for that and I will completely respect if you want to move this project into a different direction.

fdietze:
> Maybe introducing `mapTo` was not even a good idea. It does use call-by-name to accommodate for these situations. But maybe making an overload for `def apply[V](effect: IO[V])` would make more sense. It would make it more explicit.

I think this is a great realization. I like the idea!

I'd also like to make `.mapToValue(...)` a bit more unambiguous. What about `.constant(...)` or `.mapToConstant(...)`?

fdietze:
@marklister, @elyphas, @Busti, @mariusmuja, @ahjohannessen

Do you have more opinions, experiences or insights, that would help make a good decision?

marklister:
> @marklister, @elyphas, @Busti, @mariusmuja, @ahjohannessen
> 
> Do you have more opinions, experiences or insights, that would help make a good decision?

Unfortunately this one is far above my pay grade.  The little insight I can offer is that as far as possible, everything should obey the principle of least surprise:  that's gold if you're building something that real people are going to use.  We can always find a way to restore performance, and I don't think performance is that critical at the moment.

ahjohannessen:
I agree with all of what @LukaJCB said, I do not have time right now to share my thoughts. I will chime in later. Happy New Year :)

cornerman:
> I'd also like to make .mapToValue(...) a bit more unambiguous. What about .constant(...) or .mapToConstant(...)?

We do not have `mapToValue` yet. We have `apply[V](v: V)` and `apply[V](v: Observable[V])`. Or which method do you mean?

fdietze:
> We do not have `mapToValue` yet. We have `apply[V](v: V)` and `apply[V](v: Observable[V])`. Or which method do you mean?

I mean, If we make `apply` accept only an IO, we need another method to accept a constant.

cornerman:
> I mean, If we make `apply` accept only an IO, we need another method to accept a constant.

Ah yeah. Maybe we can make three overloads for the `apply` method, but we need to check how type inference works then. Or do you guys prefer more explicit method names for `Observable`, `IO` and a pure value?

elyphas:
My opinion is like @marklister, thank you :)

GRBurst:
> The little insight I can offer is that as far as possible, everything should obey the principle of least surprise: that's gold if you're building something that real people are going to use. We can always find a way to restore performance, and I don't think performance is that critical at the moment.

That is exactly what I thought and I was not aware of this problem until I found `mapTo`. But maybe this is a documentation issue - but @fdietze is working on that ;-)

> Ah yeah. Maybe we can make three overloads for the `apply` method, but we need to check how type inference works then. Or do you guys prefer more explicit method names for `Observable`, `IO` and a pure value?

If an overloaded `apply` can handle all three cases, I think that is the way to go, because the user does not have not reason about which function should be called.
However, type inference may be a bit tricky so in that case explicit functions like `mapTo` and `mapToValue` are good alternatives if the `apply` gets deprecated.

LukaJCB:
If overloading apply doesn't work with type inference, we can still explicitly prioritize by using extension methods :)

cornerman:
> If overloading apply doesn't work with type inference, we can still explicitly prioritize by using extension methods :)

@LukaJCB How would extension methods help here? Are we talking about an `implicit class` which provides the apply methods or am I misunderstanding what you mean with prioritize?

LukaJCB:
Yeah, that's what I was thinking of, not sure if it's needed

LukaJCB:
Maybe `onClick.emit(false)`? :)

elyphas:
`onClick.emit(false)`

 looks very intuitive
 :)

Busti:
I am still in the process of going down the whole FP rabbit hole, so it might not be the smartest idea to take my advice on that.
However, I really like the whole `onClick(false) --> ...` syntax and I plan to make use of it a lot throughout my project. Since I am mainly using an action driven component system I will probably end up with a lot of code that looks like this:
```scala
div(
  h1("Notification options"),
  p("Allow notifications for..."),
  button(onClick(ToggleMessageNotifications) --> sink)(
    "messages"
  ),
  button(onClick(ToggleGroupNotifications) --> sink)(
    "group messages"
  ),
  button(onClick(ToggleMentionNotifications) --> sink)(
    "mentions"
  )
)
```
But this is just an example. In reality I would probably code another component for toggle buttons.  
What I am trying to say is that `onClick.emit(false)` might add some bloat that is visually repelling for people new to the project.   

Did you also consider using something like `apply[F[_]](task: F[Unit])(implicit F: TaskLike[F])`?

Busti:
Please disregard everything I have said previously.  
I have just made the same mistake and it had me searching for 30+ minutes.

fdietze:
Such a learning makes it clear, what this is all about. :wink: 

So my current suggestion would be:

1. apply should accept only `IO[T]` and `() => T` to make it very obvious, that code will we executed when the `Emitter` triggers.

```scala
onClick(IO(...)) --> ...
onClick(() => ...) --> ...
```

2. For handling constants efficiently, but still being very explicit and obvious:

```scala
onClick.emitConstant(...) --> ...
```

What do you think?

PR #279, opened 2019-01-06
WIP: repairing dirty dom

### LINKED ISSUE #271: Unmanaged VNodes (external dom-manipulation)
In general working with third-party libs works quite good with Outwatch. You can use `onDom*` Hooks or `managedElement(elem => Cancelable(() => ...))` and integrate with other libraries using the dom element. But there is a problem when these libraries do a lot of dom-manipulation themselves.

**Example**: When using ui frameworks like semantic-ui, you might end up with `VNode`s that are moved inside the dom by semantic-ui. So `Outwatch` cannot see those changes and the virtual dom will get out of sync with the actual dom. Succeeding patch calls with snabbdom will therefore most likely fail.

We can workaround some of the issues when wrapping such nodes with `div.static(key)(<unmanaged vnode>)`. But this will still lead to errors when you want to remove this vnode again.

For proper handling, we could introduce an unmanaged vnode type. We just do the initial rendering of such VNodes and any updates coming from its observables. But like a `thunk`/`static` we would never rerender the node if it is already there and additionally will never remove the node. We would just call `onDestroy` and the user would need to handle that himself. For example, in semantic-ui, we could just `destroy` the element with jquery.

Any additional requirements or ideas?

fdietze:
The goal is to handle externally manipulated dom gracefully. This occurs when using js-libraries which manipulate the dom themselves.

Snabbdom assumes in its [patch operation](https://github.com/snabbdom/snabbdom#patch) that the **virtual dom exactly matches the real dom**. If this is not the case, the patch function might crash.

My idea is to mark the real dom as dirty. Then let outwatch detect dirty dom and repair it before the patch.

Based on #355

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,11 +21,11 @@ inThisBuild(Seq(
   )
 ))
 
+val jsdomVersion = "13.2.0"
+
 lazy val commonSettings = Seq(
   addCompilerPlugin("org.typelevel" %% "kind-projector" % "0.10.3"),
   addCompilerPlugin("com.github.ghik" % "silencer-plugin" % "1.6.0" cross CrossVersion.full),
-
-  requireJsDomEnv in Test := true,
 
   useYarn := true,
 
@@ -117,6 +117,16 @@ lazy val outwatchMonix = project
     )
   )
 
+lazy val outwatchRepairDom = project
+  .in(file("repairdom"))
+  .enablePlugins(ScalaJSPlugin, ScalaJSBundlerPlugin)
+  .dependsOn(outwatch)
+  .settings(librarySettings)
+  .settings(
+    name := "OutWatch-RepairDom",
+    normalizedName := "outwatch-repairdom",
+  )
+
 lazy val outwatch = project
   .in(file("outwatch"))
   .enablePlugins(ScalaJSPlugin, ScalaJSBundlerPlugin)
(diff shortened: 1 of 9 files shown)


cornerman:
Great work! This really helps when working with, e.g., dragging libraries where elements are moved and the VDom gets out of sync. From first testing, it seems to work. We will now try to use this in our application and see how it performs. If it shows to be working and usable, we can finalize this patch and see how we want this be merged into outwatch or snabbdom. We could consider moving it into the `patchVNode` function in snabbdom or do it in a patch callback in outwatch. Then we would not need to propagate the dirty flag to root element.

outwatch/src/main/scala/outwatch/dom/helpers/NativeHelpers.scala
```diff
@@ -1,19 +1,41 @@
...
 import scala.scalajs.js
 import scala.scalajs.js.annotation.JSBracketAccess
 
+object JSDefined {
+  // provides an extractor for js.UndefOr
+  // https://gitter.im/scala-js/scala-js?at=5c3e221135350772cf375515
+  def apply[A](a: A): js.UndefOr[A] = a
+  def unapply[A](a: js.UndefOr[A]): UnapplyResult[A] = new UnapplyResult(a)
+
+  final class UnapplyResult[+A](val self: js.UndefOr[A])
```
cornerman: for performance reasons we might want to drop this extra allocation :)

We could use `.fold` on the js.undefor instead.

cornerman:
@fdietze We should cleanup this code. Especially move it out of the snabbdom facade and then merge. We have used this code for quite some time now and it just work - eventhough repairing the dom is kind of an ugly solution :nail_care:

cornerman:
Okay! Let's get this ready for merge. I just want to move the code out of the snabbdom facade file.

cornerman:
@fdietze I have moved the repairdom functionality into its own project/package (outwatch.repairdom). But the tests are failing since some time. Any idea why?

PR #284, opened 2019-01-24
Fix typo in example in readme

zakpatterson:
Nesting example meant to show nesting, drops an inner tag level.


fdietze:
Thank you! Of course this library should be pro tomato!

I hope that in the future we can generate these examples automatically using mdoc. (#277)

zakpatterson:
Expected you to squash merge this one, so I left it as two commits. 

I just rebased the other one #285.

PR #290, opened 2019-02-26
add EmitterBuilder.ofNode and mapEmitter

cornerman:
This makes EmitterBuilders more useful. You can have an `EmitterBuilder[X, VNode]` via `EmitterBuilder.ofNode`. For VDomModifier, you can still use `EmitterBuilder.ofModifier`. Additionally, you may map over the result of an `EmitterBuilder` via `mapEmitter`.

~~We are restricting the EmitterBuilders to these two types, because we might need to handle a monix subscription when using observable-based EmitterBuilder transformations. And we can nicely handle these subscriptions for Nodes and Modifiers by using lifetime hooks. We could probably add another `EmitterBuilder.custom` method to have any type while handling the subscription yourself. But I am not sure whether this is really needed, because EmitterBuilders are usually used for Modifiers.~~

Update: I have added `EmitterBuilder.custom`:

```scala
def custom[E, R](f: ConnectableObserver[E] => R): EmitterBuilder[E,R]
```

So, `ConnectableObserver` is an observer that needs a subscription. Not every usage of an Emitterbuilders yields a subscription, but if you call `.transform` on the usage site, you will need a subscription for the observer to work. So, this interface just lets you handle this case every time. You can just run the subscription via `observer.connect(): Cancelable` which needs an implicit Scheduler.

### DIFF
--- a/outwatch/src/main/scala/outwatch/dom/helpers/EmitterBuilder.scala
+++ b/outwatch/src/main/scala/outwatch/dom/helpers/EmitterBuilder.scala
@@ -19,6 +19,7 @@ trait EmitterBuilder[+O, +R] { self =>
   def map[T](f: O => T): EmitterBuilder[T, R]
   def filter(predicate: O => Boolean): EmitterBuilder[O, R]
   def collect[T](f: PartialFunction[O, T]): EmitterBuilder[T, R]
+  def mapResult[S](f: R => S): EmitterBuilder[O, S]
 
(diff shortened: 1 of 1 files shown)


cornerman:
Ok, I am thinking about unifying the methods `ofModifier` and `ofVNode`. Maybe just have `EmitterBuilder.apply[EventType, T <: VDomModifier]`?

PR #302, opened 2019-05-08
AsyncBoundary before patch

cornerman:
Adds an `asyncBoundary` in each Observable of a VNode just before patching. Performance seems to be much better, because we leave room for events to be handled in between patching and do not back-pressure the times it takes to patch a VNode.


fdietze:
Does it fix #193?

cornerman:
> Does it fix #193?

Sadly not, but still makes sense performance wise.

PR #309, opened 2019-06-02
fix order of store emissions for observable actions

### LINKED ISSUE #289: Immediate action effects in store are applied and emitted before the action that caused them
This issue can occur when a `Reducer` with immediate effects is used with a `Store`.  

```scala
def reducer: Reducer[Int, Int] = {
  case (5, state) => (100, Some(Observable(1, 2, 3)))
  case (action, state) => (state + action, None)
}
```
When inputting a `5` into a store with the above reducer and an initial state of 0, the expected emissions would be `0, 100, 101, 103, 106`.  
The actual emissions are `0, 1, 3, 6, 100`. (or in very very rare cases `0, 101, 103, 106, 100` depending on the scheduler used)

The first approach to solve this I have made is to move the subscription to the `effects` observable outside of the `scan` operator.  
See: https://github.com/Busti/outwatch/blob/store4/outwatch/src/main/scala/outwatch/util/Store.scala#L87-L95  

The Idea behind this change is to have the effects always use the state in store that is saved after the action that caused the emissions has been processed.  
However, the effect actions still manage to, somehow overtake the original action and emit first, causing the emissions to be `0, 101, 103, 106, 100`.  

One further thing I have looked into is to use the `flatScan` operator, but it does not really help us in this case. But it might be worth checking out its implementation.

cornerman:
I have changed the way store reduces observable actions in the Store. Previously the order of actions and the resulting state would not be deterministic leading to wrong results. Now, we use `flatScan` to force an order on the store emissions.

On another note: I would love to get rid of the subscribe and cancel on the first subscription (https://github.com/cornerman/outwatch/blob/3601d6cc63a5e6ef9a522fb713acdad98e9fa4a3/outwatch/src/main/scala/outwatch/util/Store.scala#L86). It fixes the obvious problem that late observer get the correct latest value from Store. But what if I suscribe, unsubscribe, and then again subscribe. Action send to store between the unsubscribe and the second subscribe would still be lost. Any ideas what we could? Does it make sense to make the store observable hot and give a way to the user cancel the store alltogether?

@Busti Can you check, whether this fixes the problem for you?

fixes #289

### DIFF
--- a/outwatch/src/main/scala/outwatch/util/Store.scala
+++ b/outwatch/src/main/scala/outwatch/util/Store.scala
@@ -69,26 +69,17 @@ object Store {
   )(implicit s: Scheduler): IO[ProHandler[A, (A, M)]] = IO {
     val subject = PublishSubject[A]
 
-    val fold: ((A, M), A) => (A, M) = {
-      case ((_, state), action) => Try { // guard against reducer throwing an exception
-        val (newState, effects) = reducer.reducer(state, action)
-
-        effects.subscribe(
-          next => subject.feed(next :: Nil),
-          // just log the error, don't push it into the subject's observable, because it would stop the scan "loop"
-          error => dom.console.error(error.getMessage)
-        )
-
-        action -> newState
-      }.recover(recoverError.andThen(f => action -> f(state)) orElse { case NonFatal(e) =>
-        dom.console.error(e.getMessage)
-        action -> state
-      }).get
-    }
+    def fold(state: M, action: A): Observable[(A, M)] = Try {
+      val (newState, effects) = reducer.reducer(state, action)
+      effects.flatScan0(action -> newState)((actionState, action) => fold(actionState._2, action))
+    }.fold(recoverError.andThen(f => Observable.now(action -> f(state))) orElse { case NonFatal(e) =>
+      dom.console.error(e.getMessage)
+      Observable.now(action -> state)
+    }, identity)
 
     val out = subject.transformObservable(source =>
       source
-        .scan0[(A, M)](initialAction -> initialState)(fold)
+        .flatScan0[(A, M)](initialAction -> initialState)((actionState, action) => fold(actionState._2, action))
         .replay(1).refCount
     )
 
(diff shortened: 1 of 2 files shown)


Busti:
Looks good :shipit: 

Just making an observable hot does not seem to help with that problem, it might be worth asking on the monix channel.

cornerman:
Though, we need to be aware that it changes the way the store works.

Previously when emitting an `Observable[Action]` for an action, the observable of actions would be subscribed and just merged in the store subject. Now, the observable of actions is concated in the store subject - meaning no other action will be emitted until the observable of actions is completed.

Busti:
You are right, thanks for noticing that. This is **not** the desired behavior.  
The most common scenario of effects I imagine is animations. An animation is stated by an action and a second action stops the animation some time later by emitting a timed effect observable which emits an action that disables the animation.  This timeout would cause other actions to not come through until each animation is finished.  
What we need is a `mergeScan0` instead of `flatScan0`, which, unfortunately, does not exits.

Busti:
> meaning no other action will be emitted until the observable of actions is completed.

I did some more experiments on `flatScan`. [Here is the playground fiddle that I used.](https://scalafiddle.io/sf/eZUlXsU/0)  
The behavior is a little different than we anticipated.  

* `flatScan` processes all incoming data in immediately and each input produces an immediate output.
* The input is never delayed nor dropped in any way.
* For each input the reduction function, `op`, is called with the input as the second (next) argument.
* When the first input is being reduced, the accumulator is set to the initial accumulator (seed).
* The returned observables are **merged**, not concatenated.  
* When an input is processed, the accumulator contains the **latest** emission of all *merged* observables returned by each previous reduction.  
* The result of `flatScan` is a **merged** observable of all observables returned by the reduction function.  
I am sorry if you already know all of this. This writeup mostly serves as documentation for myself.  

So given the circumstances that would mean that Actions would still continue being processed even when an effects observable has not completed yet. However, since processing of effects happens an a separate `flatScan`, they will not receive the new state from the "root" `flatScan`. Thus, when they process an effect, they will use the state they were originally called with and return that down into the root observable.

Issue #310, opened 2019-06-22
renderWithStore seems to be broken


rfuerst87:
Hey Guys

Today I was playing around with the latest master branch (676f94a), trying to upgrade my app from 1.0.0-RC2. This went surprisingly well, except for getting `OutWatch.renderWithStore` to work. I just couldn't get rid of the `NoStoreException`. Either I'm doing something horribly wrong, or this is an actual bug. The following snippet (shamelessly stolen and adapted from #174) demonstrates the issue:

```
object App {

  sealed trait Action
  case object Init extends Action
  case object Add extends Action
  case object Subtract extends Action

  type State = Int
  val initialState = 0

  val reducer = Reducer[Action, State]((state, action) => action match {
    case Add => (state + 1, Observable.empty)
    case Subtract => (state - 1, Observable.empty)
  })

  def main(args: Array[String]): Unit = {
    val root = for {
      store <- Store.get[Action, State]
      root <- IO(
        div(
          button("+", onClick(Add) --> store),
          button("-", onClick(Subtract) --> store),
          p("count = ", child <-- store)
        )
      )
    } yield root

    root
      .flatMap(r => OutWatch.renderWithStore(Init, initialState, reducer, "#content", r))
      .unsafeRunSync()
  }
}
```
Any hints are very appreciated.

elyphas:
Hi, sorry, I can't help much but I think you could take a look a this examples.
https://github.com/OutWatch/outwatch/issues/283
until the people who knows better this, have time to respond

cornerman:
@rfuerst87 Thanks for opening this issue. It definitely highlights, the problems of our singleton store with its `NoStoreException`. The compiler cannot guarantee correctness here.

What happens is that you are mapping over `root`, which will access the `Store` already. This happens before `Outwatch.renderWithStore` is called. But only Outwatch.renderWithStore` creates the `Store`, therefore you get an exception if you try to use it before it is created.

So, what you should do is the following:

```scala
    val root = div(
      Store.get[Action, State].map { store =>
        VDomModifier(
          button("+", onClick(Add) --> store),
          button("-", onClick(Subtract) --> store),
          p("count = ", child <-- store)
        )
      }
    )

    OutWatch.renderWithStore(Init, initialState, reducer, "#content", root).unsafeRunSync()
```

[...]

PR #311, opened 2019-06-30
Make renderWithStore accept an IO[VNode] instead of VNode.

rfuerst87:
See discussion in #310.


cornerman:
@rfuerst87 Thank you! LGTM, though I would like to have an overload of this method for `VNode` as well, because the rest of the methods accept only `VNode` - it feels a bit more consistent.

Can you add that?

rfuerst87:
@cornerman Of course I can add an overload. Shall I add it to both classes (`Outwach` and `Store`)?

cornerman:
@rfuerst87 Good question. Imho, the method in `OutWatch` should suffice, because all methods there accept a `VNode`. For the `Store`, it should be fine as is :)

rfuerst87:
Sorry for the long delay. I was extra busy the past few days.

PR #312, opened 2019-07-03
add EmitterBuilder.combine

cornerman:


LukaJCB:
Do we have a `Monoid` on our hands? :)

cornerman:
> Do we have a Monoid on our hands? :)

Oh yes, good idea for VDomModifier and EmitterBuilder.

cornerman:
There are even more typeclasses like monads and functors, we should really add for some outwatch types that we currently lack. I have some defined in my app, but I need some time to put them in a PR.

cornerman:
@LukaJCB I have updated the PR to add `Monoid` instances for `VDomModifier` and `EmitterBuilder` as well as adding `Functor` and `Bifunctor` for `EmitterBuilder`.

Issue #308, opened 2019-05-25
problems with snabbdom-custom-props.js


elyphas:
I had a project with play but I wanted to use `sbt-revolver` and I had troubles to make it work so I started a project with `akka-http` and now I have problems with:
```scala
Bundling the application with its NPM dependencies
[error] ModuleNotFoundError: Module not found: Error: Can't resolve './snabbdom-custom-props' in '/home/elyphas/bundlerTest/manikTest/client/target/scala-2.12/scalajs-bundler/main'
```
seems that is not downloaded in the right way, 

I fix the problem manually copying the file from another project that is working.
```scala
~/bundlerTest/manik/client/target/scala-2.12/scalajs-bundler/main$ cp snabbdom-custom-props.js ~/bundlerTest/manikTest/client/target/scala-2.12/scalajs-bundler/main/
```

elyphas:
by the way I compile outwatch this way:
`outwatch/sbt compile publishLocal`
and I make reference this way:
`"io.github.outwatch" %%% "outwatch" % "0.11.1-SNAPSHOT",`

fdietze:
Looks like you are using a higher scalajs-bunder version than `0.13`. Try to downgrade and see if it works.
https://github.com/OutWatch/outwatch/pull/285#issuecomment-457802838

Either this is a scalajs-bundler regression, or they have a differnt way of including custom js code now. I didn't have time to investigate yet.

elyphas:
thank you, that's the problem.
it's solved

cornerman:
Great that it works now!

We really need to fix this problem with new scala-js-bundler versions. Maybe just include it in our snabbdom fork and be done with it.

fdietze:
> Maybe just include it in our snabbdom fork and be done with it.

I think that's the way to go.

cornerman:
This is fixed now! We have added our custom props module to our snabbdom fork.

PR #285, opened 2019-01-24
Upgrade scalajs dependency and readme references

zakpatterson:
Just started to try out this library today, trying to help out with any bumps I run into.

PR #313, opened 2019-07-09
Move Custom Props Module to Snabbdom fork 

cornerman:
As discussed in #308 #285


cornerman:
Or should we update to a scala-js 1.0 milestone?

fdietze:
scala-js 1.0 should be a different PR, and if possible, we should cross-compile between `0.6` and `1.0`

PR #324, opened 2019-08-05
Remove impurity around creation of a store

zakpatterson:
The use of Try / catchNonFatal / recover / get, is basically
a debug println. Here I simply remove this usage, and allow
the MonadError[F, Throwable] of Sync[F] to handle any thrown
exceptions.

I think the intention of the deleted code is primarily to log exceptions thrown by the passed reducer. This is the responsibility of the caller/creator of the store, and thus should be handled if necessary by the caller.

### DIFF
--- a/outwatch/src/main/scala/outwatch/util/Store.scala
+++ b/outwatch/src/main/scala/outwatch/util/Store.scala
@@ -1,6 +1,6 @@
 package outwatch.util
 
-import cats.effect.IO
+import cats.effect.{IO, Sync}
 import monix.eval.Task
 import monix.execution.Scheduler
 import monix.reactive.Observable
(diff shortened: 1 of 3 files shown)


outwatch/src/main/scala/outwatch/util/Store.scala
```diff
@@ -61,16 +58,15 @@ object Store {
    * @param recoverError An optional PartialFunctio for recovering the State when an Exception occurs in the Recuer.
```
cornerman: Can you remove this comment about the now obsolete parameter?

cornerman:
Thank you! I think, we should even this pattern to other methods returning `IO`. We should pass in effect types `F[_] : Sync` in methods like `Handler.create`, `Outwatch.renderInto` as well.

zakpatterson:
I agree about using the cats-effect type classes. I’ll have a few more PRs like this soon. Thanks!

PR #325, opened 2019-08-05
Refactor remove sink deprecation

zakpatterson:
Removes deprecated Sink and a few associated things, which were deprecated several months ago.

There are other deprecations which are not used that could be deleted. For now I'm just addressing Sink and Pipe.


zakpatterson:
@cornerman  updated branch, ready for you when you can take a look.

cornerman:
Great idea! In general I am fine with removing deprecated stuff. But we have never deprecated the `Sink` object, because we needed these methods and never came up with a better name. I think, `ObserverBuilder` is good name and might even be more intuitive than `Sink`. Though, we should keep `@deprecated val Sink = ObserverBuilder`, so the code still compiles for our users. In a later stage, we can then remove this entirely. So can you please add a redirect for `Sink`?

Otherwise, this looks good to me!

PR #326, opened 2019-08-06
Deprecate redirectMap, it's just contramap

zakpatterson:


fdietze:
Nice! Deleted code is the best code!

zakpatterson:
I do think it's instructive to look at monix's implementation of contramap, and how much more involved it is than redirectMap was. It makes me worry about the other flavors of this function in this project.

cornerman:
@zakpatterson I looked at the implementation of `contramap`, though I am not entirely sure why it is done that way.

I understand that the observer is tracking when the stream closes, has an error, or the user-code throws an exception. Then we stop in onNext, thereby enforcing the contract of observers. They should be called like this: `onNext* (onComplete|onError)?` (see https://monix.io/docs/3x/reactive/observers.html#contract).

But I am questioning why we should this. If the underlying observer already adheres to the contract, then the wrapper for `contramap` should not need to handle it as well. I am also not sure about catching the exception, because I thought it should be handled in the user-code.

Anyway, there might be reasons for this and we should probably ask in monix or open an issue to clarify.

PR #327, opened 2019-08-06
Remove implicit on Reducer constructors

zakpatterson:
I don't see any reason why these should be implicit


zakpatterson:
Sorry, now I get it. So there were implicit conversions from `(A, B) => A` to `Reducer[B, A]`.  It does make the syntax nicer for calling `create` to create a store, but I think having an implicit conversion on Function2 is odd, and users just calling `create` to make a store will probably never figure out that they can pass a Function2 in place of a Reducer object. It's a style difference though so just feel free to close if you prefer to leave it as is.

cornerman:
Good point! Generally, I am definitely in favor of removing these implicits, because implicit conversions are not really intuitive and as you said not really discoverable.

Though, I would like to iterate on the naming of the `Reducer` factory methods.

Just a proposal from the top of my head (I am totally open to other naming ideas):
- `apply` instead of `justState`
- `withEffects` instead of `stateAndEffects`
- `withOptionalEffect` instead of `stateAndOptionIO`

Additionally, for `stateAndEffects`, it should just work when passing a function, because of SAM (https://dzone.com/articles/java-8-functional-interfaces-sam).

zakpatterson:
Those names sound good. I might also try just replacing Reducer with a type alias to Function2.

zakpatterson:
I replaced the definition of reducer to be a type alias to `Function2`, and changed the names for the helper constructors as suggested by @cornerman:

> * `apply` instead of `justState`
> * `withEffects` instead of `stateAndEffects`
> * `withOptionalEffect` instead of `stateAndOptionIO`

outwatch/src/main/scala/outwatch/util/Store.scala
```diff
@@ -18,7 +18,7 @@ object Store {
    * @tparam A The Action Type
```
Busti: This `@param reducer` comment needs to be removed.

Busti: Github won't let me comment on the actual line. It is right above the one I highlighted.

zakpatterson: Thank you, fixed in 6929127

Issue #321, opened 2019-07-23
Handling of sync and async effects


cornerman:
We have this one big elephant in the room, which is missing an appropriate issue: handling of async effects. There is an issue about supporting rendering of `Task`, `Future` and `IO` (#256), but before being able to adopt this, we need to fix how we currently handle `IO`.

We interpret any `IO[T]` (with an `AsVDomModifier[T]` instance) as an `EffectModifier` (https://github.com/OutWatch/outwatch/blob/master/outwatch/src/main/scala/outwatch/dom/VDomModifier.scala#L79) and later on just render it with `unsafeRunSync()` (https://github.com/OutWatch/outwatch/blob/master/outwatch/src/main/scala/outwatch/dom/helpers/DomUtils.scala#L240). But this is a problem, because not all `IO` are meant to be just run sync (we have `SyncIO` for this intention).

So why not simply convert `IO` to `Observable`, thereby making it async and be done with it? Because, this would kill performance for any construction using `Handler.create[T]`. By wrapping the handler in `IO` and then letting the handler update the dom, we get way too many updates. For example:

```scala
val myComponent: IO[VDomModifier] = Handler.create[T].map(handler => span(...))

div(
  "Hi",
  myComponent
)
```

If we would interpret `myComponent` async, we would first render `div("Hi")`. Afterwards when the IO is finished, we would update the div again with the needed lifetime hooks for subscribing to the handler. And then we would again update the div, whenever the handler emits a value. That means making `IO` async by default, would cost us one extra update per component.

I think, we need to make the default behaviour for `IO` async, because this is the only correct way. Therefore, we need to change the signature of `Handler.create[T]` from `IO[T]` to `SyncIO[T]` and/or parametrize the create function so you can wrap the handler in any effect type you like. This will break a lot of user-code. But we could maybe write some scalafix patches to at least help with the migration.

marklister:
> This will break a lot of user-code.

We've done this before.  I think people using Outwatch pre 1.0 signed up for a moving target and are prepared to refactor their code...

cornerman:
Fixed in #355

PR #328, opened 2019-08-06
Remove overcommitment to IO in favor of typeclasses, and SyncIO instances

### LINKED ISSUE #135: Leverage cats-effect typeclasses

zakpatterson:
Fixes #135

This got pretty big, as this change touches so many places.  Sorry, I don't think there could have been an easy way to break it up.

I made a few decisions:

* `Reducer` came out of the `Store` object and resides in its own file now,
* Generally this:
```scala
object Container {
   def someMethod(arg: String): IO[Result]
}
```

Is transformed to this:
```scala
trait ContainerOps[F[_]] {
  object Container {
    def someMethod(arg: String)(implicit F: Sync[F]): F[Result]
  }
}
```
Introducing the trait there means that users can either mixin `MonixOps[F]` to get `Store`, `Handler,` `ProHandler`, etc oriented to their choice of `F`, or just `import outwatch.io._` to get the IO versions of all those things.


fdietze:
For reference, here is the old PR: https://github.com/OutWatch/outwatch/pull/178

cornerman:
@zakpatterson Thank you for all your contributions :rocket: 

I did not have time to look into this PR in detail. But just want to mention this issue as well: https://github.com/OutWatch/outwatch/issues/321, because using effect types that can be run sync would be beneficial for performance.

zakpatterson:
@cornerman thank you as well! I fully expect this one to simmer for a few days. I want to give it a good review as well.

outwatch/src/main/scala/outwatch/util/LocalStorage.scala
```diff
@@ -11,8 +11,8 @@ import org.scalajs.dom.window.{localStorage, sessionStorage}
 import outwatch._
 import outwatch.dom.dsl.events
 
-class Storage(domStorage: dom.Storage) {
-  private def subjectWithTransform(key: String, transform: Observable[Option[String]] => Observable[Option[String]])(implicit scheduler: Scheduler):IO[Handler[Option[String]]] = {
+class Storage[F[_]](domStorage: dom.Storage) extends ProHandlerOps[F] {
```
zakpatterson: I think the Sync constraint can be moved up here, so it doesn't have to be on every method.

bench/src/main/scala/Performance.scala
```diff
@@ -16,7 +17,7 @@ object jsdom extends js.Object {
   def jsdom(innerHTML: js.UndefOr[String]): js.Any = js.native
 }
 
-object Performance {
+object Performance extends HandlerOps[IO] {
```
zakpatterson: Performance could be turned into `class Performance[F[_]: Sync] extends HandlerOps[F]`, and then we could test if there are performance differences between various implementations (IO, Task, SyncIO, etc) (see #321).

README.md
```diff
@@ -108,7 +108,7 @@ object Main {
     
     val myComponent = div("Hello World")
 
-    OutWatch.renderReplace("#app", myComponent).unsafeRunSync()
+    OutWatch.renderReplace[IO]("#app", myComponent).unsafeRunSync()
```
zakpatterson: Having to specify this type parameter here each time is a little annoying. It makes me wonder if the base functions on Outwatch could be wrapped into a trait where the `F` is specified, similar to `MonixOps[F]`.

outwatch/src/main/scala/outwatch/dom/helpers/EmitterBuilder.scala
```diff
@@ -1,4 +1,5 @@
-package outwatch.dom.helpers
+package outwatch
```
zakpatterson: Revert

zakpatterson:
I think I need to add AsVDomModifier instances for Sync[F] if possible. F[VDomModifier] does not work yet. So I'll add some more instances for AsVDomModifier, and a few more cases for VDomModifier.

outwatch/src/main/scala/outwatch/dom/package.scala
```diff
@@ -1,5 +1,3 @@
 package outwatch
 
-package object dom extends Implicits with ManagedSubscriptions with MonixOps {
-  val Handler = outwatch.Handler
-}
+package object dom extends Implicits with ManagedSubscriptions
```
zakpatterson: Need to restore here so ProHandler is available in the dom package as before.

outwatch/src/main/scala/outwatch/dom/helpers/EmitterBuilder.scala
zakpatterson: Nevermind, this is actually good because the ObserverOps are in the outwatch package.

bench/src/main/scala/Performance.scala
zakpatterson: This will be addressed as a separate PR, too many things change due to use of unsafeRunSync in bunches of places inside this object.

README.md
zakpatterson: This works now.

zakpatterson:
Ok, at this point existing code should compile with a few deprecation warnings.

zakpatterson:
I'm getting pretty close on this. The current hang up is how to create AsVDomModifier instances for given `F`s that the user wants.

So background:

As a means to get syntax like `div( ........)` to work, `div.apply` accepts any number of `VDomModifier`s.  Many things can be implicitly converted to `VDomModifier`, such as `IO[VDomModifer]`, `Obserable[VDomModifier]`, `Observable[IO[Observable[Seq[IO[VDomModifier]]]]]`, [etc](https://github.com/OutWatch/outwatch/blob/0049b5a4c74acc6d0edf1b7152862499dafeec9c/outwatch/src/main/scala/outwatch/AsVDomModifier.scala#L17)

So all we need to do is create an AsVDomModifier instance for any new effect types we need.  But looking at [the instance for IO](https://github.com/OutWatch/outwatch/blob/eb34ffd03f2eb5461d2babb2473953d62fd742a8/outwatch/src/main/scala/outwatch/dom/VDomModifier.scala#L79), it's not so easy:

```final case class EffectModifier(effect: IO[VDomModifier]) extends VDomModifier

**Comment by cornerman** [comment]
I do not have that much time, but I'll just try to give a response.

> And this bit is troubling in a few ways. One it means we can't extend this pattern for Fs constrained only by Sync because Sync provides no means of running. Also just having unsafeRunSync called this often seems like it would be problematic.

This call of `unsafeRunSync` is the core of the issue in #321. It is definitely wrong, because even for `IO`, we cannot and should not assume that we can run it synchronously in javascript. We can assume it for `SyncIO`. Apart from that, it is okay and not called multiple times. The effect is only run once for your VNode and only ever after you run your `OutWatch.renderInto`.

For the `EffectModifier`, I would propose making it really a modifier for synchronous effects: `SyncEffectModifier(unsafeRun: () => VDomModifier) extends VDomModifier`. For effect types that can and should (?) be run sync, we can add an extra `AsVDomModifier` instance, for example for `SyncIO`. I would then handle other effects types asynchronously with `Observable.from`.

Now, in the issue #321, I have added some reasons why handling all your effects async could degrade performance. We will need to find a way to allow for the desired flexibility without hard overhead.

What do you think?

**Comment by cornerman** [comment]
> Many things can be implicitly converted to VDomModifier, such as IO[VDomModifer], Obserable[VDomModifier], Observable[IO[Observable[Seq[IO[VDomModifier]]]]], etc

It is really sad that we need to convert implicitly for `VDomModifier`, but currently really needed for the apply method...

**Comment by cornerman** [comment]
`monix` has a typeclass `TaskLike`, which you might be able to use for converting multiple different effect types to `Observable`.

**Comment by zakpatterson** [comment]
So just to summarize where we are, and incorporate some gitter talk:

Currently we've removed the dependency of IO in a ton of places, allowing the user to specify their own effect context for the OutWatch.render* methods, and in *defining* dom elements.

However, the interpretation of this F only exists at the moment for IO. We've identified reasons above why IO is a bad choice, but switching to a different monad for this would have been hard until now because there was so much dependency on it.

So for now, we're going to switch the critical section of code that actually manages `VDomModifier`s, the core of the dom algebra we're working with, to use SyncIO instead of IO, because of its better guarantees in javascript.

>For the EffectModifier, I would propose making it really a modifier for synchronous effects: SyncEffectModifier(unsafeRun: () => VDomModifier) extends VDomModifier. For effect types that can and should (?) be run sync, we can add an extra AsVDomModifier instance, for example for SyncIO. I would then handle other effects types asynchronously with Observable.from.

This ^^^ is next.

> monix has a typeclass TaskLike, which you might be able to use for converting multiple different effect types to Observable.

I think @cornerman means `ObservableLike` here, and bad implicit conflicts happen at the moment if this is used instead of Observable, which I can cover in a different PR. This project also introduces an `AsObservable` typeclass which I think should go away and be replaced with `ObservableLike`. `ObservableLike` is different though, it extends `FunctionK`, not commonly used for typeclasses, so maybe it will work, or maybe there's something left I don't understand.

Ok here comes the SyncIO part next, then we should be ready to wrap this one.

**Comment by zakpatterson** [comment]
I think that's it. There are now instances for SyncIO and IO, and users can build their programs more generically, delaying specification of their effect type until their main method.

Other stuff that could be part of this PR but we're leaving as next steps:
* Documentation of this feature is necessary, as part of #276, #277
* Revising the unsafeRunSync() stuff #321

**Comment by fdietze** [comment]
This PR be tested with this jitpack dependency:
```
"com.github.zakpatterson" % "outwatch" % "61c0787"
```

**Comment by zakpatterson** [comment]
> This PR be tested with this jitpack dependency:
> 
> ```
> "com.github.zakpatterson" % "outwatch" % "61c0787"
> ```

Now `"com.github.zakpatterson" % "outwatch" % "447ec4d"`

Only update needed for user code should be:

```
import outwatch.util._
import outwatch.util.io._ 
``` 

instead of just the first import, and

```
import outwatch.dom._
import outwatch.dom.io._
```
instead of just the first.

File: outwatch/src/main/scala/outwatch/dom/Handler.scala [hunk 6]
```diff
@@ -1,34 +1,28 @@
...
+import monix.reactive.observers.Subscriber
+import monix.reactive.subjects.{BehaviorSubject, ReplaySubject}
 import monix.reactive.{Observable, Observer}
 
 import scala.concurrent.Future
 
-trait MonixOps {
-  type ProHandler[-I, +O] = Observable[O] with Observer[I]
-  type Handler[T] = ProHandler[T,T]
+trait HandlerOps[F[_]] {
```
cornerman: My problem is that I am not sure which effect type we should choose by default for `Handler.create` (this is about #321). If we want to handle `IO` async in the future, we will have one extra patch wherever we use `IO`. For `SyncIO`, we can do it synchronously, which is much better for constructing the `VNode` for snabbdom patching, because the initial patch contains everything after the effect is run. Whereas async effects can only be patched in afterwards.

What I am asking is whether it makes sense to have `Handler.create[F, T]` instead of `trait Handler[F[_]] { def create[T] }`? I think, it can be useful for mixing different effect types in the same app.

outwatch/src/main/scala/outwatch/dom/AsValueObservable.scala
```diff
@@ -28,4 +28,3 @@ object AsValueObservable extends AsValueObservableInstances0  {
     }
   }
 }
-
```
cornerman: not needed ;)

outwatch/src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -76,6 +76,7 @@ final case class NextVDomModifier(modifier: StaticVDomModifier) extends StaticVD
 case object EmptyModifier extends VDomModifier
 final case class CompositeModifier(modifiers: js.Array[_ <: VDomModifier]) extends VDomModifier
 final case class ModifierStreamReceiver(stream: ValueObservable[VDomModifier]) extends VDomModifier
+final case class EffectModifierIO(effect: SyncIO[VDomModifier]) extends VDomModifier
```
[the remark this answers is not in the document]
cornerman: I would even like to have something like `VDomModifier.delay { impure code }`, it could be named `VDomModifier.effect` as well.

outwatch/src/main/scala/outwatch/dom/helpers/STRef.scala
```diff
@@ -1,18 +1,19 @@
 package outwatch.dom.helpers
 
-import cats.effect.IO
+import cats.effect.Sync
+import cats.implicits._
 
-class STRef[A](private var unsafeGet: A) {
-  def put(a: A): IO[A] = IO { unsafeGet = a; a }
+class STRef[F[_]: Sync, A](private var unsafeGet: A) {
```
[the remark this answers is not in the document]
zakpatterson: I don't have a strict preference one way or the other. I agree though about requiring the least necessary constraint, so perhaps `getOrThrow` could only require `Monad` as written, and maybe `put` could be rewritten only requiring `Applicative`. But if we remove the constraint at the class level, then you could write `val ref = new STRef[List, Int]`, and end up with an instance that has no callable methods. In this case the choice of F seems intertwined with the functionality of the class itself, and I think should be constrained at the class level.

But that said, this class should actually be deleted in its entirety.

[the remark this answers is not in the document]
cornerman: Fair enough, let's just leave it as is then.

outwatch/src/main/scala/outwatch/dom/AsValueObservable.scala
zakpatterson: Fixed in f64e66c

outwatch/src/main/scala/outwatch/dom/Compat.scala
```diff
@@ -1,27 +1,29 @@
...
 
+import cats.effect.SyncIO
 import monix.execution.Scheduler
 import org.scalajs.dom.{ClipboardEvent, DragEvent, KeyboardEvent, MouseEvent}
 import outwatch.dom.helpers.{ChildStreamReceiverBuilder, ChildrenStreamReceiverBuilder}
 
-trait Handlers {
-  @deprecated("Use Handler.create[MouseEvent] instead", "0.11.0")
+trait Handlers extends HandlerOps[SyncIO]{
+  @deprecated("Use Handler.create[F, MouseEvent] instead", "0.11.0")
```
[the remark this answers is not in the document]
zakpatterson: Fixed in 0ab26e6

outwatch/src/main/scala/outwatch/dom/Handler.scala
```diff
@@ -1,34 +1,28 @@
...
-    def redirectCollect[I2](f: PartialFunction[I2, I]): Observer[I2] = redirectMapMaybe(f.lift)
-    def redirectFilter(f: I => Boolean): Observer[I] = redirectMapMaybe(e => Some(e).filter(f))
+    def unsafe[T]: Handler[T] = ReplaySubject.createLimited(1)
+    def unsafe[T](seed:T): Handler[T] = BehaviorSubject[T](seed)
   }
+}
+
+trait ProHandlerOps[F[_]] extends HandlerOps[F] {
 
   implicit class RichProHandler[I,O](self: ProHandler[I,O]) {
```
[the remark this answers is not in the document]
zakpatterson: Fixed in 1cf9400. The `ProHandler.create` method did require a context, but `ProHandler.apply` and ProHandler.connectible did not. It appears to me that `.apply` and `.connectable` are internal/private api methods, so I put those methods in a separate private object. Let me know if you think these are methods users would need, but I think all users would create via `ProHandler.create`.

outwatch/src/main/scala/outwatch/dom/helpers/STRef.scala
```diff
@@ -1,18 +1,19 @@
...
+  def put(a: A): F[A] = Sync[F].delay { unsafeGet = a; a }
 
-  def getOrThrow(t: Throwable): IO[A] = IO(unsafeGet)
-    .flatMap(s => if (s == null) IO.raiseError(t) else IO.pure(s)) // scalastyle:ignore
+  def getOrThrow(t: Throwable): F[A] = 
+    unsafeGet.pure[F].flatMap(s => if (s == null) Sync[F].raiseError(t) else s.pure[F]) // scalastyle:ignore
```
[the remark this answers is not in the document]
zakpatterson: I'm matching the behavior of the prior implementation, but unsafeGet is just a value at this point correct?  Maybe the class definition should have a by-name parameter.

outwatch/src/main/scala/outwatch/dom/helpers/STRef.scala
```diff
@@ -1,18 +1,19 @@
...
-  def update(f: A => A): IO[A] = IO { unsafeGet = f(unsafeGet); unsafeGet }
+  def get: F[A] = getOrThrow(new IllegalStateException())
+  def update(f: A => A): F[A] = Sync[F].delay { unsafeGet = f(unsafeGet); unsafeGet }
 }
 
 object STRef {
-  def apply[A](a: A): STRef[A] = new STRef(a)
-  def empty[A]: STRef[A] = new STRef[A](null.asInstanceOf[A]) // scalastyle:ignore
+  def apply[F[_]: Sync, A](a: A): STRef[F, A] = new STRef[F, A](a)
+  def empty[F[_]: Sync, A]: STRef[F, A] = new STRef[F, A](null.asInstanceOf[A]) // scalastyle:ignore
```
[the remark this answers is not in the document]
zakpatterson: Cool, updated this in e011b9a

outwatch/src/main/scala/outwatch/dom/Handler.scala
zakpatterson: I think it makes sense, the only problem I have with moving the `F` into the method is that it breaks type inference. Since `Handler.create` doesn't accept an argument with an `F`, it means you'll always have to specify the full return type as a type argument or between `:` and `=`.  This might even be desirable in this case, not sure.

outwatch/src/main/scala/outwatch/dom/helpers/EmitterBuilder.scala
```diff
@@ -1,7 +1,8 @@
...
+package dom.helpers
```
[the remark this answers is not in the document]
zakpatterson: `ObserverOps` are part of the outwatch package. So the alternative is to `import outwatch._`. This feels cleaner to me, but I'm happy to do it either way.

outwatch/src/main/scala/outwatch/dom/Handler.scala
zakpatterson: Another big point here is that it doesn't really make sense for `Handler.create` or `ProHandler.create` to return inside of `F`. All these methods are doing is `Sync[F].delay {  ... }` where nothing in `...` cares about `F`. So we could simply leave it up to the caller to create the handler inside of an F if needed.

zakpatterson: I pushed up a commit that looks at removing F from ProHandler and Handler, 

https://github.com/zakpatterson/outwatch/commit/54db1632b955a052c884b48b04cf7e8320a4ded7

I also refactored all the test code in this commit (painful), to show how that would look. I thought I'd get your feedback before merging that into this PR. It does seem a bit simpler to me at least, to just have the `Handler.create` etc return the Handler without wrapping in `F`.

outwatch/src/main/scala/outwatch/util/LocalStorage.scala
```diff
@@ -45,20 +46,20 @@ class Storage(domStorage: dom.Storage) {
...
-  def handler(key: String)(implicit scheduler: Scheduler): IO[Handler[Option[String]]] = {
+  def handler(key: String)(implicit scheduler: Scheduler): F[Handler[Option[String]]] = {
     val storageEvents = storageEventsForKey(key)
     subjectWithTransform(key, Observable(_, storageEvents).merge)
   }
 }
 
-object LocalStorage extends Storage(localStorage)
-object SessionStorage extends Storage(sessionStorage)
+class LocalStorage[F[_]: Sync] extends Storage[F](localStorage)
```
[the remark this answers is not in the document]
zakpatterson: The way it is currently, Storage depends on ProHandlerOps which needs a decision on what F we're in. I address this in my other comment, [my commit here](https://github.com/zakpatterson/outwatch/commit/54db1632b955a052c884b48b04cf7e8320a4ded7) also solves this issue.

outwatch/src/main/scala/outwatch/dom/VDomModifier.scala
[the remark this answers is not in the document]
zakpatterson: I'm having a hard time with the `.unsafeRunSync()` call anywhere in the library code. I hope there's a better way.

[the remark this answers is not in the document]
zakpatterson: This PR is getting too big though, dealing with unsafeRunSync belongs in a subsequent one.

outwatch/src/main/scala/outwatch/dom/helpers/STRef.scala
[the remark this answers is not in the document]
cornerman: Afaik, before we have delayed constructor with `IO(unsafeGet)`, which will read the var at the point the IO is run. Now with `unsafeGet.pure[F]`, we read the var at the time the method is called.

outwatch/src/main/scala/outwatch/dom/Handler.scala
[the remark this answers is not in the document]
cornerman: Nice! I think, you can make the `ProHandlerBuilder`, because user might want to extend on our custom monix ops, where there might need to construct a `ProHandler`. 

How about naming this object `ProHandler` again and then add extension methods to `ProHandler` in the `ProHandlerOps`? We could do the same for `Handler`, so have the basic `unsafe` method at least.
```
implicit class ProHandlerMethods(factory: ProHandler.type) { def create[T] = ... }
```

outwatch/src/main/scala/outwatch/dom/Handler.scala
cornerman: That is an excellent point to bring up. The advantage of enforcing an effect for `Handler.create` is that users are forced to building referentially transparent components/functions when handling state in the form of `Handler` in your app.

Take for example:

```scala
def counter = Handler.create[Int].map(handler => div(handler, onClick(handler.map(_ + 1)) --> handler)
div(counter, counter) // you get two separate counter components
```

Now, we have a method without wrapping in an effect called `Handler.unsafe` which is at least explicit in the naming. 

```scala
def counter = {
  val handler = Handler.unsafe[Int]
  div(handler, onClick(handler.map(_ + 1)) --> handler)
}
div(counter, counter) // you get two dependent counter components that share the same state
```

Apart from that, some users just might not care about these properties and like to just handlers without IO. So your argument to wrap it yourself is actually quite good, it kind of becomes more flexible. But the question is whether it is explicit enough.

@LukaJCB What do you think?

outwatch/src/main/scala/outwatch/dom/helpers/EmitterBuilder.scala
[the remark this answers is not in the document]
cornerman: I am just not used to these "appended" package definitions, but it is fine with me :)

outwatch/src/main/scala/outwatch/dom/VDomModifier.scala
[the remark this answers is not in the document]
cornerman: Okay, let's leaves this part out of this PR then. I think user code normally does not depend on `EffectModifier`/`EffectModifierIO`, so we can just rename/redefine it subsequently.

outwatch/src/main/scala/outwatch/dom/Handler.scala
zakpatterson: Handler.unsafe is just an alias for this: https://github.com/monix/monix/blob/v3.0.0-RC3/monix-reactive/shared/src/main/scala/monix/reactive/subjects/ReplaySubject.scala#L214. Why should that even be considered unsafe?

outwatch/src/main/scala/outwatch/dom/helpers/STRef.scala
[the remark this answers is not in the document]
zakpatterson: Yep! That makes sense now, just not used to vars.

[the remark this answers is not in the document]
zakpatterson: `Applicative[F].pure` is call by value, and `Sync[F].delay` is call by name, so that's why it behaves this way.

outwatch/src/main/scala/outwatch/dom/Handler.scala
cornerman: As, I tried to describe in my previous comment, defining the method `Handler.create` without IO would not be referential transparent. It returns a stateful interface (`Observer`).

If I would define create like:

```scala
def create[T] = PublishSubject[T]
div(counter(create[Int]), counter(create[Int]))
```

then it is not equivalent  to:

```scala
def create[T] = PublishSubject[T]
val handler = create[Int]
div(counter(handler), counter(handler))
```

zakpatterson: With IO and without, you would still have two different handlers.  I think in both cases those two examples are not equivalent.

cornerman: Without `IO`, you do not have different handler (if you use the return value twice instead of calling the function twice). With  `IO`, it is equivalent, because in both cases you get a different handler.

zakpatterson: That’s right, sorry. thanks.

zakpatterson: Whereas with iO if you wanted two counters showing the same state, you would map over IO and pass the inner value twice.

zakpatterson: Thanks for pushing back on this one. Another concern I had was the type inference issue, but I solved that with the partial application trick.  

So now you don't have to write `Store.create[IO, Action, Model](intiialAction, model)`. You can just write `Store.create[IO](initialAction, model)`. Same with `Handler.create`.  See  
d80a55b.

outwatch/src/main/scala/outwatch/util/LocalStorage.scala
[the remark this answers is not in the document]
zakpatterson: Done in 3551279

outwatch/src/main/scala/outwatch/dom/Handler.scala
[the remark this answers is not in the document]
zakpatterson: At this point, given the other comments, we've deleted ProHandlerOps, so I think we can just revert the ProHandlerBuilder

outwatch/src/main/scala/outwatch/dom/helpers/STRef.scala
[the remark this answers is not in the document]
zakpatterson: Done in fc06528. Also found a method to get rid of that flatMap.

outwatch/src/main/scala/outwatch/dom/Handler.scala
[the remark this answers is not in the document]
zakpatterson: Reverted it in 02d41b0.

zakpatterson:
Ok @cornerman, I just made a pass to try to minimize this changeset a bit.

outwatch/src/main/scala/outwatch/util/LocalStorage.scala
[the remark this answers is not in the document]
zakpatterson: Further in a9c928b

outwatch/src/main/scala/outwatch/dom/helpers/EmitterBuilder.scala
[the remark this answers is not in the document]
zakpatterson: This has been reverted since I've minimized the changeset, this is not needed.

outwatch/src/main/scala/outwatch/util/Store.scala
```diff
@@ -1,64 +1,29 @@
...
-     */
-    def apply[A, M](f: (M, A) => M): Reducer[A, M] = (s: M, a: A) => f(s, a) -> Observable.empty
-
-    /**
-     * Creates a Reducer with an optional IO effect.
-     */
-    def withOptionalEffects[F[_]: ObservableLike, A, M](f: (M, A) => (M, Option[F[A]])): Reducer[A, M] = { (s: M, a: A) =>
-      f(s, a).map(_.fold[Observable[A]](Observable.empty)(Observable.from))
-    }
+  final class CreatePartiallyApplied[F[_]](val dummy: Boolean = false) extends AnyVal {
```
cornerman: Alternatively, instead of a dummy val, we could try `@inline class` and `@inline def` and check whether the scala-js compiler optimizes it properly.

outwatch/src/main/scala/outwatch/util/package.scala
```diff
@@ -0,0 +1,13 @@
...
+import monix.reactive.Observable
+
+package object util {
+  /**
+   * A Function that applies an Action onto the Stores current state.
+   * @param reducer The reducing function
+   * @tparam A The Action Type
+   * @tparam M The Model Type
+   */
+  type Reducer[A, M] = (M, A) => (M, Observable[A])
```
cornerman: Why did we move the `Reducer` out of the `Store` object? To me, it felt right to be namespaced with the Store, because this is not any Reducer, but the Reducer for the Store.

outwatch/src/main/scala/outwatch/util/Store.scala
cornerman: I have tested this in fastOpt and inlining does reliably remove the overhead of this partial application trick (see https://github.com/OutWatch/outwatch/pull/343/commits/1d4df805bb92cb6eb298f1642d299df56b64fa99).

For reference, the code generated for `Handler.create[IO, Int](0)` is:
```javascript
  $m_Loutwatch_Handler$().create__O__Lcats_effect_Sync__O(0, $m_Lcats_effect_IO$().ioEffect$2);
```

The generated code for `Handler.create[IO](0)` as it is now with AnyVal:
 ```javascript
$m_Loutwatch_Handler$CreatePartiallyApplied$().apply$extension__Z__O__Lcats_effect_Sync__O(false, 0, $m_Lcats_effect_IO$().ioEffect$2);

// this $m_Loutwatch_Handler$CreatePartiallyApplied$-method somehow constructs a singleton for CreatePartiallyApplied
var $n_Loutwatch_Handler$CreatePartiallyApplied$ = (void 0);
function $m_Loutwatch_Handler$CreatePartiallyApplied$() {
  if ((!$n_Loutwatch_Handler$CreatePartiallyApplied$)) {
    $n_Loutwatch_Handler$CreatePartiallyApplied$ = new $c_Loutwatch_Handler$CreatePartiallyApplied$().init___()
  };
  return $n_Loutwatch_Handler$CreatePartiallyApplied$
}
```

The generated code for `Handler.create[IO](0)` with `@inline` annotations:
 ```javascript
  var F = $m_Lcats_effect_IO$().ioEffect$2;
  $m_Loutwatch_Handler$().create__O__Lcats_effect_Sync__O(0, F);
```

The last snippet generates an extra variable for the implicit `Sync` instance, but nothing else.

cornerman: So, I think for this PR it is good as is. We will work on some part in another PR.

outwatch/src/main/scala/outwatch/Handler.scala
```diff
@@ -9,27 +10,37 @@ import monix.reactive.{Observable, Observer}
...
 
 object Handler {
-  def empty[T]:IO[Handler[T]] = create[T]
+  def empty[F[_]: Sync, T]: F[Handler[T]] = create[F, T]
 
-  def create[T]:IO[Handler[T]] = IO(unsafe[T])
-  def create[T](seed:T):IO[Handler[T]] = IO(unsafe[T](seed))
+  def create[F[_]] = new CreatePartiallyApplied[F]
+  def create[F[_], T](implicit F: Sync[F]): F[Handler[T]] = F.delay(unsafe[T])
+  def create[F[_], T](seed: T)(implicit F: Sync[F]): F[Handler[T]] = F.delay(unsafe[T](seed))
```
cornerman: I am wondering, whether we should additionally give users a default effect to use with `Handler.create`. So they do not need to think which effect they should use in each case. The current generic ones are nice, when you want to mix them with other effects in a for-comprehension. But if you are just doing `Handler.create[Int].map(handler => div(handler, ...))`, you do not really need to worry which effect type is used -- defaulting to SyncIO here would be a good choice performance-wise.

Should we maybe have another set of methods where you do not need to provide the effect type, for example:

```scala
def apply[T] = create[SyncIO, T]
def apply[T](seed: T) = create[SyncIO, T](seed)
```

Not sure about the naming, whether it should be `apply` or maybe `create` and `createF`. What do you think?

zakpatterson:
Passing the ball to @cornerman with #343

cornerman:
Thank you for this beautiful PR! On top of the functionality, this really enables us to finally fix handling of async effects.

My initial intention was actually to merge your PR first and then merge my PR later. But superseding is also fine now, because I think, it basically works.

outwatch/src/main/scala/outwatch/Handler.scala
cornerman: Any opinions? :)

PR #334, opened 2019-08-17
Remove singleton store and unused global STRef class

zakpatterson:
The deleted code provides a singleton store. I propose we remove it, as it's untested, and users can just hold onto their own reference to the store they create.


cornerman:
Nice! Though we need to have an agreement, how we handle deprecations and removing code. This will not be backwards-compatible and I am unsure what approach we should take:

- Keep deprecations until we finally have some stable version and can release 1.0. This will become more and more tedious.
- Make one more RC release with a lot of deprecations and point migrating users to that release, so they can fix the deprecations. Then they can update where these APIs will be removed.
- Just move fast and accept that we are not really backwards-compatible until 1.0.

Not sure which approach to take.

zakpatterson:
I think most of the deprecation warnings were added a long time ago. Why don't we do a RC with all the deprecated code deleted, and all deprecation warnings deleted, and then move forward from that point with a goal of responsible deprecation/backwards compatibility thru 1.0?

cornerman:
I have talked about backwards-compatibility and how to move forward with @fdietze as well. We agree that we want to keep the basic API compatible, so that most stuff keeps on working. But we also want to move forward with OutWatch, so if some parts of the API will break that is accepted for now. So for the next RC release, we want to get rid of deprecations and further improve on our API (especially with regards to #328).

zakpatterson:
As a newcomer, with a tiny project started on this library, I'm fully behind removing all deprecated code and moving forward!

PR #338, opened 2019-08-21
update help

elyphas:
Just a simple line to see how is going.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,7 +344,7 @@ div("Argh!", bigFont)
 if you want to reuse it but overriding a property:
 
 ```scala
-VDomModifier(bigFont, fontSize := "99px")
+val bigFont2 = VDomModifier(bigFont, fontSize := "99px")
 ```
 
 You can also use a `Seq[VDomModifier]` directly instead of using `apply` defined in the [VDomModifier](src/main/scala/outwatch/dom/package.scala) object.
--- a/README.md
+++ b/README.md
@@ -326,6 +326,7 @@ val vnode: VNode = div()
 val modifiers: List[VDomModifier] = List("Hello", id := "main", color := "tomato", vnode)
 ```
 
+
 Every `VNode` contains a sequence of `VDomModifier`. A `VNode` is a `VDomModifier` itself.
 
 There are implicits for converting primitives, `Option[VDomModifier]`, `Seq[VDomModifier]` to `VDomModifier`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -340,6 +341,12 @@ div("Argh!", bigFont)
 // <div style="font-size: 40px; font-weight: bold;">Argh!</div>
 ```
 
+if you want to reuse it but overriding a property:
```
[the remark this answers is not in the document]
fdietze: @elyphas, after you accept the change (click the button under the suggestion), I will merge this PR.

fdietze:
@elyphas I got your changes, thank you. Will review them soon.

elyphas:
Thank You!

README.md
```diff
@@ -425,6 +432,84 @@ div(
...
+
+We have `Observables` and `Observers` from monix.
+
+and `Handler` from Outwatch:
+
+Handler is just an alias for `Observer with Observable` and is referential transparent API. (before was `Sink with Observer` where `Sink` was just a wrapper around Observer). 
+
+So `Handler` creates an `Observable` that is also a `Observer`.
+
+Older version  patterns changes from.
```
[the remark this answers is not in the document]
fdietze: Yo can remove the lines from the readme, commit again and push to this branch. Then it will be removed from this PR.

README.md
```diff
@@ -425,6 +432,84 @@ div(
...
+val additions = createHandler[Int]()
+```
+
+to 
+
+```scala
+val additions = Handler.create[Int].unsafeRunSync()
+```
+
+When is the right time to put the .unsafeRunSync on there?
```
fdietze: It should be noted here, that the whole idea of the handler wrapped in IO is **referential transparency**. If the component is in a `def`, it is already referential transparent, because for each invocation a new `Handler` is created, so the usage `unsafeRunSync()` is fine in this case (@cornerman correct me if I'm wrong). If the same code is stored in a `val` and the `val` is used several times, the created `Handler` is hared between the usages and is therefore not referential transparent.

cornerman: The whole idea of handler being wrapped in an effect type is indeed referential transparency. Though, it does not matter whether the component is in a `def` or a `val`, you should not run `unsafeRunSync` within your app, but in the best case only once in your main, when running `OutWatch.render*`. You can map over `IO[Handler]` if you want to share some handler state, but it will be explicit.

Take for example:

```scala
def counter = Handler.create[Int].map(handler => div(handler, onClick(handler.map(_ + 1)) --> handler)
val counterResult = counter

// the following two line are equivalent:
div(counter, counter)
div(counterResult, counterResult)
```

Now, if you run unsafeRunSync in your counter component, your method will not be referentially transparent:

```scala
def counter = {
  val handler = Handler.unsafe[Int]
  div(handler, onClick(handler.map(_ + 1)) --> handler)
}
val counterResult = counter

// the following two line are NOT equivalent:
div(counter, counter)
div(counterResult, counterResult)
```

README.md
```diff
@@ -340,6 +341,12 @@ div("Argh!", bigFont)
 // <div style="font-size: 40px; font-weight: bold;">Argh!</div>
 ```
 
+If you want to reuse it, but want to overwrite a property:
```
cornerman: Maybe we can even be more specific. How about this?

"If you want to reuse the `bigFont`, but want to overwrite one of its properties, you can combine two `VDomModifier`. Here the latter `fontSize` will overwrite the one from `bigFont`:"

PR #340, opened 2019-08-22
[WIP] Scala 2.13 upgrades

zakpatterson:
This is pending a non-snapshot release of Monix 3, and merge/release of https://github.com/raquo/scala-dom-types/pull/52.


build.sbt
```diff
@@ -6,12 +6,16 @@ inThisBuild(Seq(
 
   organization := "io.github.outwatch",
 
-  scalaVersion := "2.12.9",
+  scalaVersion := "2.12.8",
```
fdietze:
```suggestion
  scalaVersion := "2.12.9",
```

build.sbt
```diff
@@ -45,11 +49,11 @@ lazy val outwatch = project
     normalizedName := "outwatch",
 
     libraryDependencies ++= Seq(
-      "io.monix"      %%% "monix"       % "3.0.0-RC3",
+      "io.monix"      %%% "monix"       % "3.0.0-eddc79d-SNAPSHOT",
```
fdietze: would RC5 work here now? https://github.com/monix/monix/releases

outwatch/src/main/scala/outwatch/AsVDomModifier.scala
```diff
@@ -26,14 +25,14 @@ object AsVDomModifier {
   }
 
   implicit def arrayModifier[T : AsVDomModifier]: AsVDomModifier[Array[T]] = (value: Array[T]) =>
-    CompositeModifier(value.map(VDomModifier(_))(breakOut) : js.Array[VDomModifier])
-
+    CompositeModifier(js.Array(value.map(VDomModifier(_)).toIndexedSeq : _*))
```
fdietze: looks expensive to me. `map` produces a new `Array`, then this array is passed as varargs into `js.Array()` which afaik produces a defensive copy in 2.13, which is then copied again into a `js.Array`...

zakpatterson: We need to find some intermediate ground though, as breakout is gone in 2.13

https://www.scala-lang.org/blog/2017/02/28/collections-rework.html#views

We could do a bifurcated source tree, I don't have a lot of time for this right now though.

build.sbt
zakpatterson: Nope: 

https://oss.sonatype.org/content/repositories/public/io/monix/

outwatch/src/main/scala/outwatch/dom/helpers/Builder.scala
```diff
@@ -24,7 +24,14 @@ object AttributeBuilder {
...
   protected def name: String
-  @inline def accum(s: String): AccumAttrBuilder[T] = accum(_ + s + _)
+  @inline def accum(s: String): AccumAttrBuilder[T] = new AccumAttrBuilder[T](
+    name, 
+    this, 
+    {
+      case (v1, v2) => s"$v1$s$v2"
+    }
+  )
+
```
cornerman: Why was this code changed? I kind of liked the older version better :)

zakpatterson: see 217803

zakpatterson: I suppose it could have been `_.toString + _.toString`?

outwatch/src/main/scala/outwatch/dom/ChildCommand.scala
```diff
@@ -79,7 +79,7 @@ object ChildCommand {
           children.prepend(VNodeProxyNode(SnabbdomOps.toSnabbdom(node)))
         case ReplaceAll(list) =>
           children.clear()
-          children.push(list.map(node => VNodeProxyNode(SnabbdomOps.toSnabbdom(node))): _*)
+          children.push(list.map(node => VNodeProxyNode(SnabbdomOps.toSnabbdom(node))).toSeq: _*)
```
cornerman: Instead of mapping and converting, we can do `list.foreach(children.push(...))` directly instead.

outwatch/src/main/scala/outwatch/dom/DomTypes.scala
```diff
@@ -103,9 +103,11 @@ trait HtmlAttrs
     new PropBuilder(key, codec.encode)
 
   // super.className.accum(" ") would have been nicer, but we can't do super.className on a lazy val
-  override lazy val className = new AccumAttrBuilder[String]("class",
-    reflectedAttr(attrKey = "class", propKey = "className", attrCodec = codecs.StringAsIsCodec, propCodec = codecs.StringAsIsCodec),
-    _ + " " + _
+  override lazy val className = new AccumAttrBuilder[String](
+    "class",
+    reflectedAttr(attrKey = "class", propKey = "className", attrCodec = codecs.StringAsIsCodec, propCodec = codecs.StringAsIsCodec), {
+      case (v1, v2) => s"$v1 $v2"
```
cornerman: This might be controversial and normally I like the `PartialFunction` for Tuple better, but here I would prefer to keep the old version. To me, it is more readable.

outwatch/src/main/scala/outwatch/AsVDomModifier.scala
cornerman: I agree with both of you, some potentially expensive copying was introduced with regards to collections. But we also need to find some ground for supporting 2.13.

What is a "bifurcated source tree"?

outwatch/src/main/scala/outwatch/dom/helpers/Builder.scala
cornerman: Oh, okay. I guess that is even a good thing. So nevermind!

outwatch/src/test/scala/outwatch/SnabbdomSpec.scala
```diff
@@ -7,9 +7,13 @@ import outwatch.dom.OutWatch
 import outwatch.dom.dsl._
 import cats.effect.IO
 import snabbdom._
+import cats.effect.ContextShift
+import scala.concurrent.ExecutionContext.Implicits.global
 
 class SnabbdomSpec extends JSDomAsyncSpec {
 
+  implicit val cs: ContextShift[IO] = IO.contextShift(global)
```
cornerman: Why does this test need a `ContextShift` now?

outwatch/src/main/scala/outwatch/AsVDomModifier.scala
zakpatterson: Just that it is possible to have for example a `src/main/scala_2.12` directory that provides alternate sources if compiling for scala 2.12.

cornerman: Ah, thanks! Now i know what you mean. I hope, we can avoid splitting the source tree, but let's see. I think, we just need to got through all the instances here where a copy might be introduced and check whether we can do something about it.

outwatch/src/test/scala/outwatch/SnabbdomSpec.scala
cornerman: It is needed for `IO.fromFuture`.

outwatch/src/main/scala/outwatch/dom/helpers/DomUtils.scala
```diff
@@ -222,7 +223,7 @@ private[outwatch] object NativeModifiers {
           lengthBefore += lengthsArr(i) getOrElse 1
           i += 1
         }
-        modifiers.splice(lengthBefore, lengthsArr(index) getOrElse 1, mods.toSeq: _*)
+        modifiers.asInstanceOf[js.Dynamic].splice.asInstanceOf[FunctionRawApply].applyCall(modifiers, js.Array(lengthBefore, lengthsArr(index) getOrElse 1).concat(mods))
```
cornerman: This call might look weird on the first glance, but was only way I was able to properly call `splice` on an Array. Normally `splice` takes a `VarArg*` for the value that should be inserted. Now with defensive copying and toSeq, we always get boilerplate code:

```

cornerman:
I have added a commit to resolve the discussed pain points with regards to intermediate collections being created.

PR #341, opened 2019-08-22
add combine method to VDomModifier

cornerman:


outwatch/src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -11,7 +11,9 @@ import snabbdom.{DataObject, VNodeProxy}
 import scala.scalajs.js
 import scala.scalajs.js.JSConverters._
 
-sealed trait VDomModifier
+sealed trait VDomModifier {
+  @inline def combine(modifier: VDomModifier): VDomModifier = VDomModifier(this, modifier)
```
zakpatterson: Given the monoid instance, method syntax is already available via `import cats.syntax.monoid._`. Why also have it here?

cornerman: True, you can get it from cats already. But I was also asked by people whether there is a `combine` method -- so this would be for users that are new to `cats` so it is easily discoverable for everybody. My motivation is that everybody just gets this method autocompleted in their IDE without needing any imports or knowledge about cats.

fdietze: hmm. I understand both points. In general I would like to avoid this duplicated API. How many newcomers ask for a `combine` method?

zakpatterson: Implementing a subset of cats to avoid people having to use cats while we're using cats and they're already getting cats transitively as a dependency seems wrong. Where would that end.

zakpatterson: Also I noticed this project is in typelevel incubator, so I think it would only be right to embrace that ecosystem fully.

cornerman: Okay, makes sense.

Though, we might want to add similar methods like `prepend` or `append` to VDomModifier, so something similar will eventually be added to this trait. But you are right, we should not duplicate the cats API. Thanks!

PR #343, opened 2019-09-01
[WIP] Sync and Async Effects

cornerman:
Based on #328


cornerman:
@LukaJCB What do you think about this typeclass [RunSyncEffect](https://github.com/OutWatch/outwatch/blob/1d4df805bb92cb6eb298f1642d299df56b64fa99/outwatch/src/main/scala/outwatch/dom/RunSyncEffect.scala)? It is kind of the equivalent of `cats.Async` just for running some effect types sync. And afaik `cats.SyncIO` is exactly that: "SyncIO is similar to IO, but does not support asynchronous computations. Consequently, a SyncIO can be run synchronously to obtain a result via unsafeRunSync." (from https://typelevel.org/cats-effect/datatypes/syncio.html).

We can now render any `cats.Effect` type and not only `IO`, but we can only do so async and not sync as we did before (which was wrong). For performance reasons, we still want to be able to inject effect into a `VNode` that are run sync. This is useful so that the initial patch can already contain synchronously run effects. For convenience, I have added a `VDomModifier.delay { ...modifier }` method to easily create an effectful computation of a VDomModifier. But that is not really an effect type and you cannot do anything with. So making use of `SyncIO` would make sense (having `RunSyncEffect` as a way to extend this).

cornerman:
@zakpatterson What do you think about these changes? I took up some parts we have discussed based on your PR

LukaJCB:
@cornerman I like it!  
One thing I can think of to improve would be to separate the running from the interpreting. 
How about something like this:
```scala
trait RunSyncEffect[F[_]] {

  def interpret[G[_]: Sync]: PartiallyApplied[G]

  class PartiallyApplied[G[_]: Sync] {
    def apply[A](fa: F[A]): G[A]
  }
}

//then you can call it like this
SyncEffect[F].interpret[SyncIO](fa)
```

outwatch/src/main/scala/outwatch/AsVDomModifier.scala
```diff
@@ -86,6 +88,19 @@ object AsVDomModifier {
   implicit def effectRenderAs[F[_] : cats.effect.Effect, T : AsVDomModifier]: AsVDomModifier[F[T]] = (effect: F[T]) =>
     ModifierStreamReceiver(ValueObservable(Observable.fromTask(Task.fromEffect(effect).map(VDomModifier(_)))))
 
+  implicit object FutureRender extends AsVDomModifier[Future[VDomModifier]] {
+    def asVDomModifier(future: Future[VDomModifier]): VDomModifier = future.value match {
```
zakpatterson: interesting!

cornerman:
@LukaJCB Oh cool, that is a really nice encoding! Would something like that make sense to be included in cats-effect? Would be really useful.

I am a bit torn between the two solutions just because of boring performance reasons. If this was to be used in a more generic library, I would definitely do it as you propose. Just because in our code this is a hot-spot when patching and we could save one or two allocations when running it directly :)

LukaJCB:
Yeah, this should really be part of cats-effect in the next version, I'll make sure of it ;)

PR #344, opened 2019-09-01
add append method to VNode (apply is alias of append)

cornerman:


outwatch/src/main/scala/outwatch/dom/VDomModifier.scala
```diff
@@ -82,30 +82,36 @@ final case class StringVNode(text: String) extends VDomModifier
...
   def apply(args: VDomModifier*): BasicVNode
+  def append(args: VDomModifier*): BasicVNode
   def prepend(args: VDomModifier*): BasicVNode
   def thunk(key: Key.Value)(arguments: Any*)(renderFn: => VDomModifier): ThunkVNode = ThunkVNode(this, key, arguments.toJSArray, () => renderFn)
   def thunkConditional(key: Key.Value)(shouldRender: Boolean)(renderFn: => VDomModifier): ConditionalVNode = ConditionalVNode(this, key, shouldRender, () => renderFn)
   @inline def thunkStatic(key: Key.Value)(renderFn: => VDomModifier): ConditionalVNode = thunkConditional(key)(false)(renderFn)
 }
 final case class ThunkVNode(baseNode: BasicVNode, key: Key.Value, arguments: js.Array[Any], renderFn: () => VDomModifier) extends VNode {
-  def apply(args: VDomModifier*): ThunkVNode = copy(baseNode = baseNode(args))
+  @inline def apply(args: VDomModifier*): ThunkVNode = append(args: _*)
```
fdietze: we have to verify if inlining works reliably here. Especially in 2.13 which makes a defensive copy afaik.

fdietze: Is it possible somehow to make a real alias?

cornerman: Agree, the aliasing is quite manual, having to define it in every instance, because I need the correct return type.

I have verified inlining manually in `2.12` (don't know how it looks like in `2.13`). It does inline correctly, so `append` and `apply` are compiled into the same code.

I have additionally added the inline annotation to the `VNode` classes, so if you write div.apply(1).apply(2).apply(3). We will concat three arrays, but do not construct the intermediate VNode instances.

fdietze: Ok. I think verification for `2.12` is good enough for now.

cornerman: It would be great to have some kind of testing that inlining works correctly :)

PR #345, opened 2019-09-02
special EventObservable for dom.Event with sync event operations

### LINKED ISSUE #227: PreventDefault, StopPropagation helpers on window/document events

cornerman:
This allows `stopPropagation`, `preventDefault` and `stopImmediatePropagation` on global events from window or document.

We have these event helpers for constructing observables out of dom events on window/document. We now return a special `EventObservable[dom.Event]` instead, which extends Observable and adds additional methods for event operations. Old code will work just the same.

```scala
dsl.events.document.onClick : EventObservable[dom.Event]
dsl.events.window.onStorage : EventObservable[dom.Event]
dsl.events.document.onClick.stopPropagation.preventDefault
dsl.events.window.onStorage.stopImmediatePropagation
```

Additionally I have updated the code to adhere to the `Ack` of the underlying `Observer`. If we receive an `Ack.Stop`, we will now unregister the event listener and stop feeding the stopped observer.

fixes #227


outwatch/src/main/scala/outwatch/dom/EventObservable.scala
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    register()
+
+    Cancelable(() => unregister())
+  }
+
+  private def withOperator(newOperator: EV => Unit): EventObservable[EV] = new EventObservable(target, eventType, { ev => operator(ev); newOperator(ev) })
+
+  def preventDefault: EventObservable[EV] = withOperator(_.preventDefault)
```
fdietze: A test to verify the behavior (especially sync operations) would be nice. I find it difficult to see the correctness from only looking at the code.

cornerman: I have added a basic test for the Observable itself and one test having a sync operation (stopPropagation). The test does not really assure that the operation is done sync, but it at least assures that the values in the observable did go through the operation.

PR #346, opened 2019-09-02
Remove hfunction from Snabbdom Facade

cornerman:
And update snabbdom fork to 0.7.5. Contains latest changes from snabbdom/snabbdom master-branch and remove parsing of selector tags to find encoded `#id.cls` in selector tags of snabbdom (we do not need this).

I have moved the snabbdom fork to our organization now: So the new dependency is from outwatch/snabbdom. The old dependency will keep on working because of githubs redirections.

Based on #355

### DIFF
--- a/bench/yarn.lock
+++ b/bench/yarn.lock
@@ -3638,9 +3638,9 @@ signal-exit@^3.0.0:
   resolved "https://registry.yarnpkg.com/signal-exit/-/signal-exit-3.0.2.tgz#b5fdc08f1287ea1178628e415e25132b73646c6d"
   integrity sha1-tf3AjxKH6hF4Yo5BXiUTK3NkbG0=
 
-"snabbdom@git://github.com/cornerman/snabbdom.git#semver:0.7.4":
-  version "0.7.4"
-  resolved "git://github.com/cornerman/snabbdom.git#2814da5d929d432b9c3fc5ecf91ca03c72e17bbe"
+"snabbdom@git://github.com/outwatch/snabbdom.git#semver:0.7.5":
+  version "0.7.5"
+  resolved "git://github.com/outwatch/snabbdom.git#f92761a40914825d7b8f702f607f9d929f95033f"
 
 snapdragon-node@^2.0.1:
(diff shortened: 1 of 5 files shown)


outwatch/yarn.lock
```diff
@@ -2,3929 +2,538 @@
 # yarn lockfile v1
```
fdietze: why remove it?

cornerman: Oh, that was not my intention.

cornerman: It is not removed, but much smaller. I do not know why though :D

outwatch/src/test/scala/outwatch/OutWatchDomSpec.scala
```diff
@@ -33,6 +33,18 @@ class OutWatchDomSpec extends JSDomAsyncSpec {
     elem.dispatchEvent(event)
   }
 
+  def newProxy(tagName: String, dataObject: DataObject, string: String) = new VNodeProxy {
```
fdietze: maybe inline these factories?

outwatch/yarn.lock
fdietze: where is it then?

cornerman: It is gone :)

outwatch/src/test/scala/outwatch/OutWatchDomSpec.scala
cornerman: It's just a test :)

outwatch/yarn.lock
cornerman: It seems like the require `snabbdom/h` pulled that all in? If add hfunction back in, I get the old huge lock file, without the hfunction it gets smaller :parasol_on_ground:

outwatch/src/test/scala/outwatch/OutWatchDomSpec.scala
fdietze: Oh right. Nevermind. :+1:

PR #351, opened 2019-09-05
Really skip publishing bench project

cornerman:
The current master mistakenly publishes the bench project, such that our jitpack dependencies break (because there are now multiple artifacts for outwatch in jitpack).

Thanks to @Busti for pointing this out!


cornerman:
I will be merging this without further approval, because jitpack dependencies to master are currently broken.

Issue #354, opened 2019-09-07
Remove implicit to VDomModifier


cornerman:
Currently, we have an implicit conversion to `VDomModifier` for every type that has an `AsVDomModifie` instance: https://github.com/OutWatch/outwatch/blob/master/outwatch/src/main/scala/outwatch/dom/Implicits.scala#L10.

This is pretty much only to allow using `AsVDomModifier` instances with variadic arguments. So, you can write `div(1, "string")` and have the `Int`/`String` be converted to `VDomModifier` automatically for you. But this has the disadvantage that you get this conversion everywhere when just using the type `VDomModifier` yourself: `val x: VDomModifier = IO { VDomModifier(...) }`. It might be useful in cases but I do not like these magic conversions.

I would like to propose the following:

```scala
// This is a trick for allowing methods accepting variadic args to use a
// typeclass instance.
// We want to write a method `def foo(modifiers: VDomModifier*)`, which
// accepts every type that has an AsVDomModifier instance. For methods
// accepting only one or a sepcific number of arguments, we can just require
// an AsVDomModifier instance with `def foo[T: AsVDomModifier](mod: T)`.
// But there are no variadic type args, so we need an implicit conversion,
// though we do not want to have a global implicit to VDomModifier. Therefore
// we invent a super type of T called `Converted[T]` to which we then
// have a global implicit -- it does not trigger when requiring a normal
// T because it is a super type.  So you can write the desired function
// like `def foo(modifiers: Converted[VDomModifier]*)` without
// affecting the usage of the VDomModifier type itself.

private[outwatch] object Tagging {
  type Converted[T] >: T

  @inline def to[T](mod: Converted[T]): T = mod.asInstanceOf[T]
  @inline def toF[F[_], T](mod: F[Converted[T]]): F[T] = mod.asInstanceOf[F[T]]
}

```

Then we can define:

```scala
object VDomModifier {
  type Converted = Tagging.Converted[VDomModifier]
} 
```

Then just define our methods with variadic args as:
```scala
def apply(modifiers: VDomModifier.Converted*) = ???
```

What do you think?

cornerman:
Additional it would be good to get rid of two implicit conversions we still have:

- https://github.com/OutWatch/outwatch/blob/master/outwatch/src/main/scala/outwatch/helpers/Builder.scala#L27
- https://github.com/OutWatch/outwatch/blob/master/outwatch/src/main/scala/outwatch/package.scala#L8

[...]

PR #347, opened 2019-09-02
[WIP] Refactoring EmitterBuilder

cornerman:
based on #328

Reduces the generated code size, because abstraction like SyncEmitterBuilder did lead to runtime cost before. Additionally this should be much more readable. 

Some changes:

- `EmitterBuilder[O,R]` is now a type alias to `EmitterBuilderExecution[O, R, Execution]`
- `EmitterBuilder.Sync[O,R]` instead of `SyncEmitterBuilder[O,R]` (is a type alias for `EmitterBuilderExecution[O, R, SyncExecution]`
- `Execution`/`SyncExecution` type param marks whether the current `EmitterBuilder` runs sync and therefore whether sync methods like `stopPropagation`/`preventDefault` can be called
- Rename `EmitterBuilder.apply[E]("event")` to `EmitterBuilder.fromEvent[E]("event")`
- Rename `EmitterBuilder.ofNode[E](f)` to `EmitterBuilder.apply[E,VNode](f)`
- Rename `EmitterBuilder.ofModifier[E](f)` to `EmitterBuilder.apply[E,VDomModifier](f)`
- Rename `EmitterBuilder.custom[E,R](f)` to `EmitterBuilder.apply[E,R](f)`

probably fixes #295 (TODO: needs test)

PR #355, opened 2019-09-08
FRP agnostic

### LINKED ISSUE #96: Make Outwatch FRP-backend agnostic

cornerman:
# FRP Agnostic OutWatch

Support other streaming libraries via typeclasses. More or less minimal internal reference implementation for the streaming typeclasses. Library support like monix needs one more import. Most of the new code resides in [`colibri`](https://github.com/cornerman/colibri).

# Todos
- [x] update readme
- [x] fix tests and bugs
- [x] move monix to different sub-project
- [x] move reactive to different sub-project
- [x] migration path
- [x] decide on naming (Still please feel free to comment on the naming, input is very welcome)
- [ ] ~~instances for third-party libraries: airstream, scala.rx, fs2, ...~~ (should be done in another PR)
- [ ] update template (will be done after merging this PR)

## Streaming Typeclasses

Relevant for streaming, managed subscriptions and emitter:
- `Sink[F[_]]` can send values and errors into `F` has an `onNext` and `onError` method.
- `Source[F[_]]` can subscribe to `F` with a `Sink` (returns a cancelable subscription)
- `CanCancel[T]` can cancel `T` to stop a subscription
- `SubscriptionOwner[T]` can let type `T` own a subscription
- `LiftSink[F[_]]` can lift a `Sink` into type `F`
- `LiftSource[F[_]]` can lift a `Source` into type `F`
- `SubscriptionOwner[T]` that a type T can own and manage a subscription

We have some concrete types and a very basic reference implementation for `Sink` and `Source`:
- `Observer[T]`
- `Observable[T]`
- `Subject[A]` / `ProSubject[I,O]`
- `Cancelable`

Instances for monix are provided as well. The reference implementation is built with the js-runtime in mind - and not suitable for the JVM.

Methods like `subscribe`, `onNext`, `onError` in the streaming typeclasses are not referential transparent, we should probably use an effect type there. But this was out-of-scope for this PR - it would have made this refactoring more difficult. In the end, it is the same as before, the subscribe methods on monix observables are also not referential transparent. For the most part, users do not interact with this API anyhow, but use higher-level concept in our dsl that are referential transparent.

Should there be laws for these typeclasses? Like handlers should keep their last value and implementations should do that.

## Rendering

There are `Render[T]` (previously called `AsVDomModifier`) instances for any `Source` and you can write to any `Sink` in your `EmitterBuilders`. So you can just use them.

## Handler factories

We had `Handler.create*` before and we keep it. We have two typeclasses for creating handlers:
-  `CreateHandler[F[_]]`
- `CreateProHandler[F[_,_]]`

We have descriptions of how factories for handlers, sources and streams look like:
- `SinkFactory[F[_]]`
- `SourceFactory[F[_]]`
- `HandlerFactory[F[_]]`

Now we stick all these together with `LiftSource` and `LiftSink` and we get a `HandlerEnvironment`, which has types aliases and factories for `Handler`, `ProHandler`, `HandlerSource`, `HandlerSink` for whatever types you are using. We just need some instances of the above typeclasses. The default handler environment is using our own implementation is imported via `import outwatch.reactive.handler._` giving you all Handler factories and types:

```scala
import outwatch.reactive.handler._ // or outwatch.reactive.handlers.monix._

val handler: SyncIO[Handler[Int]] = Handler.create[Int](1)
val source: HandlerSource[Int] = HandlerSource[Int](3)
val sink: HandlerSink[Int] = HandlerSink.create(onNext, onError)
...
```

You typically just want one of these Environments in scope, the types would name-clash overwise. And the handler environment is totally optional, you can create your Handlers of the library of your choice yourself and everything will work out of the box without the environment. The environment is just to have a vocabulary of how to create sources, sinks, handlers without using concrete types names from third-party libraries. This way you can switch a whole file to another streaming library by merely changing an import from `outwatch.reactive.handler._` to `outwatch.reactive.handlers.monix._`. Of course, this will never cover all complex details that might be needed, but is target for basic use-cases and should be enough for most things. You have the streaming typeclasses as well to abstract further over your types.

## EmitterBuilder improvements

Cleaner interface and new methods like:
- `mapSync(e => synceffect)`
- `concatMapAsync(e => effect)`
- `concatMapFuture(e => future)`
- `transform[F[_] : Source](f: SourceStream[?] => F[?])`
- Deprecate `onClick(value)` in favor of `onClick.use(value)` (fixes #275 ).
- `useSync` and `useAsync` for using an effect.
- And many more... 

Per default transformations are done with our own source implementation, but you can `lift` into another type. And e.g. map your `EmitterBuilder` events as an `Observable` from monix.

Additionally `forwardTo` is now an alias for `-->` and there is another method without arguments to just subscribe to the emitter without doing something with the values. Useful for effects, e.g.: `onClick.concatMapAsync(s => Task(...)).discard`.

## What has changed?

Deleted types: `AsObserver`, `AsValueObservable`, `ValueObservable`, `ObserverBuilder`. Removed deprecations. Some minor types are moved like most notably `AsVDomModifier` (now called `outwatch.Render`) and some others - should not be user-facing.

The package `dom` was removed, all import are renamed `outwatch.dom.*` to `outwatch.*`.

The EmitterBuilder type has moved from the `helpers` package to the main package (`outwatch.EmitterBuilder`).

In the EmitterBuilder some functions were deprecated and we had to rename to functions:  `useLatest` is now `useLatestEmitter` and `withLatest` is now `withLatestEmitter`.

Imports have changed a bit. Like before this is enough for basic stuff:

```scala
import outwatch._ // as before, but a bit smaller: VDomModifiers, AsVDomModifier, managed subscriptions, child commands,
import outwatch.dsl._ // as before the dsl from domtypes
```

You can get support for monix observables and observers with an optional import that you now need extra if working with monix:
```scala
import colibri.ext.monix._
```

The above is missing Handler or ProHandler types and factories, you get these through an `HandlerEnvironment`. There is a default one with our very simple implementation which works well for basic stuff (`outwatch.reactive.handler`). And there is one for monix, the one we had before (`outwatch.reactive.handlers.monix`). It is entirely optional just for creating reactive stuff in a library agnostic way:
```scala
import outwatch.reactive.handler._ // or outwatch.reactive.handlers.monix._
```

Due to separting out monix, there is not first-class support for a `Scheduler` anymore. If you are working with monix, you will need an implicit `Scheduler` from time to time when building you VNode (needed for some implicit resolutions of instances). But as before, nothing is subscribed as long as you do not render and subscriptions are cleaned up just as before.

Another notable change: the stuff in `outwatch.util` and `outwatch.http` was moved to its own project `outwatch.util`. It depends on the colibri reactive library instead of monix. You can use by depending on `outwatch-util`.

We have helpers for global events in the dsl like `events.window.onStorage`. These do now return a `Observable`, our own implementation, instead of a monix observable. You can still lift the `Observable` into any `Source` type if you want: `events.window.onStorage.lift[monix.reactive.Observable]` or just work with the `Observable`: `events.window.onStorage.map(g).foreach(f)`.

The `AttributeBuilder` can now be used with an `assign` method. Before, you could only write `cls := "some-class"`, now you can additionally write `cls.assign("some-class")` as an alias.

## Compatability

*First of all: there will be incompatibilities and user-code will break. Most of the times this can be fixed with imports and some minor renaming.*

Splitting monix out of the outwatch core and abstracting over the streaming part, made some breaking changes inevitable. I tried to keep them to a minimum, but also wanted to take the opportunity to cleanup some code. So generally outwatch components should just compile - but some imports will need to be changed and minor renamings are needed in some places.

Thanks to @zakpatterson, we can now leverage the cats.Effect typeclasss (See #328 and #343). This change already made some minimal API changes.

Now taking these two changes combined, I think this is also a good point in time to cut some old ties -- that is remove deprecated stuff. So, I took the liberty to remove all deprecations in this PR, which makes the code a bit simpler.

I have further made some minor changes (most notably in the last commit), where some packages are moved (they are public but normally do not occur in user-code).

## Migration Path

- Add a dependency to `outwatch-monix`
- Add a dependency to `outwach-util`
- Rename all imports `outwatch.dom._` to `outwatch._`
- Rename import `outwatch.helpers.EmitterBuilder` to `outwatch.EmitterBuilder`
- Add two imports to every file (you can delete unused ones later):
```scala
import colibri.ext.monix._
import outwatch.reactive.handlers.monix._ // or if you want to use our own reactive library instead of monix: import outwatch.reactive.handler._
```
- If you depend on monix operations like `transformHandler` or `mapHandler`, they are called `transformSubject` or `mapSubject` and you need to import: `import colibri.ext.monix.ops._`
- Replace usages of `.useLatest` to `.useLatestEmitter` and rename `.withLatest` to `.withLatestEmitter`.
- Replace usages of `outwatch.AsVDomModifier` with `outwatch.dom.Render`.
- Replace usages of `Sink.fromFunction` or `Sink.create` with `Observer.create`.
- Wherever you are using `outwatch.dom.dsl.events.?`, you now get a `Observable` instead of a monix `Observable`. For most cases it should just compile, if you need an `Observable` instead, for example: `outwatch.dom.dsl.events.window.onStorage.lift[monix.reactive.Observable]`.
- Methods like `OutWatch.renderInto` or `OutWatch.renderReplace` now take a type argument for the effect type, for example: `OutWatch.renderInto[IO](...)` or `OutWatch.renderReplace[SyncIO](...)`.
- Creating a store with `Store.create` now takes a type argument for the effect type, for example: `Store.create[IO](...)` or `Store.create[SyncIO, ...](...)`.
- The handler factory `Handler.create` now returns `SyncIO` instead of `IO`. You can provide any other `cats.effect.Sync` type as well and get for example `IO` again: `Handler.createF[IO](...)`.
- Review your usage of `IO`. As of now, all effects are rendered *async*. If you want to render effects that are run synchronously, you can use `SyncIO`. Another alternative to delay the execution of a modifier is: `VDomModifier.delay(...)`.

## Performance

Performance seems to be faster than the current master when using our internal implementation  (~2.57 times faster). When using monix instead of our internal one, we also see an improvement in performance (~1.87 times faster). We should take these numbers with a grain of salt, because the benchmarks do not give a complete picture about the performance in a web application. Results of our `ChildrenPerformance` benchmarks in nanoseconds:
```
// master
Children      273412185
Thunk          41723023
Command        21165999

// This PR: colibri
Children       99377714
Thunk          29386730
Command        16145791

// This PR: monix
Children      146140733
Thunk          32927143
Command        18532105
```

I am currently experimenting with a trick to get values synchronously out of some streams by doing a premature subscription. When we know we are going to be rendered but are not rendered yet, we can already subscribe and update the modifiers.Thereby saving a patch for the initial value of the involved streams which would otherwise come in after the initial rendering. Especially monix seems to profit from this optimization. Then the results look like this:
```
// This PR: colibri (premature subscription)
Children      106122394
Thunk          29528723
Command        15799843


// This PR: monix (premature subscription)
Children      152536458
Thunk          32615416
Command        18623817
```

With the latest updates premature-subscriptions are now consistently a bit slower than subscribing at mount-time (at least in our benchmarks) :/ I do not know why and my intuition is exactly the opposite. I have verified that premature subscriptions lead to less patch-calls - therefore I will keep it, eventhough the pure numbers from the benchmarks tell a different story. 

But there is also one disadvantage: if you are nesting handlers in one element, you will now always get a second patch for the inner handler. Before all observables in one node were flattened internally and therefore this could not happen. Example:
```scala
// map over a handler and nest an inner handler inside as a VDomModifier
handler.map { _ =>
  VDomModifier(innerHandler)
} 
// map over a handler and switchMap the inner handler
handler.switchMap { _ =>
  innerHandler
} 
``` 

Previously these two examples behaved pretty much the same, because the observables were flattened for you -- but this also had overhead because we needed an extra step to map through all observables. Now with this PR, these examples are not necessarily equivalent. When the outer handler in the first example triggers, the inner observable is not yet subscribed and will only be subscribed after the first patch and then rendered in a second patch. We have some of these constructions in the tests, you can see where I had to change some render-counters (some increased, some decreased). I still think, this is acceptable. This might be happening sometimes because of our global implicit to VDomModifier, but we will try to fix this in #354.

Based on #343 (so ignore the first commit for reviewing changes).

Fixes #96


cornerman:
Does anyone know why this does not compile on 2.11? This was already happening in the #347 PR (see travis).

cornerman:
I have prepared code for the VDomModifier conversion (#354) in this PR. But did not enable it, because then even more code would break when users update. Maybe better to do deliver compile error in increments :)

cornerman:
I have moved the monix instances and helpers into another project `outwatch-monix`. Due to the outwatch tests partly depending on the monix implementations, there is now another `tests` project, where most of the tests reside. They do not belong in the monix project as they do indeed test outwatch core.

elyphas:
What about akka-streams?

cornerman:
> What about akka-streams?

Afaik akka does not exist for scala-js, so that would not work right now.

elyphas:
Yes it is, well I think,

elyphas:
Before I came here, I made some good things with akka.js actors, but I also like monix, I know  monix from outwatch.
http://akka-js.org/

cornerman:
Oh, thanks for the link. I thought akka-js was unmaintained - but great if it is available. Then, it is probably possible to provide an integration for it!

cornerman:
I think, this branch is ready! If anyone of you have time, please try it out - it would help a lot. Feedback is definitely welcome :)

The jitpack dependencies are:
```
    "com.github.cornerman.outwatch" %%% "outwatch" % "0a470538"
    "com.github.cornerman.outwatch" %%% "outwatch-monix" % "0a470538"
```

If there are no show-stoppers, this PR will be merged soon.

busti:
So far this branch is doing great for me. I do not really have an opinion on the design side of things, other than liking it all. I feel like this is mostly free of any obvious bugs, so much so that it can be merged. If anything comes up, a new issue should do.

cornerman:
Great to hear! Thank you.

> I feel like this is *mostly* free of any obvious bugs

Now I am interested in the "mostly" ;)

elyphas:
Thank you
I will try soon

busti:
> Now I am interested in the "mostly" ;)

I haven't found any yet. This isn't javascript, so we are probably not going to have any `NaN` anywhere, but you never know™

cornerman:
> but you never know™

So true! I thought, you might have seen something.

reactive/src/main/scala/outwatch/reactive/Subscription.scala
```diff
@@ -0,0 +1,75 @@
+package outwatch.reactive
+
+import cats.Monoid
+
+import scala.scalajs.js
+
+sealed trait Subscription {
+  def cancel(): Unit
```
busti: I must admit that I prefer the name `Cancelable`. It is a bit more broad.

cornerman: Great Point! Subscription might be to specific. I even made an alias in `outwatch.dom` called `cancelable(() => ...)`.

I was not sure what to do with names and I am open to changes! I wanted to work in files that use `monix` and `outwatch.reactive` at the same time - having no name clashes made that easier. The same actually applies to `SinkSourceHandler`, `SourceStream` and `SinkObserver` - they could very well be named `Observable` and `Observer`. I even though about naming subscriptions `Cancellable`, but not sure.

What would you prefer?

cornerman:
This PR makes `Handler.create` return `SyncIO`. The motivation was making it explicit and being able to render the result synchronously - which we were only able to do for `SyncIO`. But now, we are actually able to render things like `IO.pure(1)` or `IO(2)` or `IO(1).map(...)` synchronously as well - things are only going async if you do something async in your IO like `IO.shift *> IO(1)` or `IO.sleep(1.mins) *> IO(2)`.

So, the question is: Should we let `Handler.create` return `IO` again?

cornerman:
Furthermore, I'd like to change the names of the concrete reactive classes we have:
- `SourceStream` to `Stream` or `Observable`
- `SinkObserver` to `Receiver` or `Observer`
- `SinkSourceHandler` to `DuplexHandler` or `Subject`
- `Subscription` to `CancelToken` or `CancellationToken` or `Cancelable`

If you like certain names more than others or have an idea for a better name, please let me know -- I really cannot decide and now we can still change it :)

On the one hand, reusing the vocabulary from rx probably makes sense, on the other hand having different names made it easier for me to work with monix and these classes in one file. What do you prefer?

For, the streaming typeclasses, the naming is `Sink` and `Source`.

elyphas:
`> On the one hand, reusing the vocabulary from rx probably makes sense, on the other hand having different names made it easier for me to work with monix and these classes in one file. What do you prefer?`

For me It would be good to keep the names from rx. I have found help on internet that help me. 

on the other hand.

What do think about the integration with akka-stream?

because in a future this could be easier to dive into the streaming concepts through `outwatch`, like I doing,  and learning on client and server at the same time.

Monix works on client and server.

also Akka-stream.

So, I think It would be better what's more easy for develop a good wrapper.

Maybe some annotations relating the names with rx help in certain way.


and to be precise in names:

 `SourceStream`  --> Source (because the source is always a stream or I think so, I think Stream, is redundant)

`SinkObserver` to `Receiver` or `Observer` (Receiver would be a completly a new name too much, maybe `Sink` or `SinkObserver` is good)

`SinkSourceHandler` to `DuplexHandler or `Subject` (`SinkSource` maybe describe better what is going to do but `Subject` is good because stick on names of rx)

`Subscription` to `CancelToken` or `CancellationToken` or `Cancelable` (
sorry I am confuse with the names in the same context, "If You are talking about to subscribe to a stream").
I do not have a clear concept of this, but I think is better `Subscription`; because is describing what is going to do and if it is cancelable is only a second feature.
We are going to subscribe to a stream and in the future we could cancel the subscription, `Cancelable` for me is just a mean to stop what we started to do.)

busti:
> So, the question is: Should we let `Handler.create` return `IO` again?

Why is `createF` not the default?  

> Furthermore, I'd like to change the names of the concrete reactive classes we have  

I'd go with `Observable`, `Observer`, `Subject` and `Cancelable`. Though I do not like it, it is the most widely adopted naming scheme, which is more understandable for new users.

cornerman:
Thank you both for your input, @busti and @elyphas!

From what I am hearing, I think we will be going for the rx names then, so it will become `Observable`, `Observer`, `Subject` and `Cancelable`.

> Why is createF not the default?

The first reason why I wanted to have a default effect type here is that in most cases I do not really care which effect type is used in `Handler.create`. My main use-case is writing simple things like `div(Handler.create[Int].map { handler => ??? })` without specifying an effect type. So when just using the handler in a `VDomModifier`, there is not much that I would gain if i chose my own effect type. On the other hand, `createF` becomes useful if I combine this with other effectful things, so you can directly work in the desired effect type - but I am not sure that this is the most common use-case.

The second reason was performance, I wanted to chose an effect type that can be run synchronously per default - because it can be rendered directly without needing to handle the possibility of it being async. Maybe I promised a bit too much too early in my last message about running IO and SyncIO - because as of now, `SyncIO` is still less overhead than rendering an `IO` made of sync operations.

Why would you prefer `createF` to be the default?

busti:
> Why would you prefer `createF` to be the default?  

I'm thinking, when we already have an implicit `F[_]` in scope most of the times, why not make that the default.  
But I might be projecting too much from my own code.

busti:
Could `ChildCommand` maybe also accept the generic streaming type?  
I would like to use the interface with `monix.reactive.Observable`

cornerman:
@busti Yes, good point! That should possible and i think it would be better that way. I will update branch, until then you can do `SourceStream.lift(observable)`.

reactive/src/main/scala/outwatch/reactive/Sink.scala
```diff
@@ -0,0 +1,17 @@
+package outwatch.reactive
+
+trait Sink[-F[_]] {
+  def onNext[A](sink: F[A])(value: A): Unit
+  def onError[A](sink: F[A])(error: Throwable): Unit
```
busti: @cornerman In order to be functionally pure, the `onNext` and `onError` in `Sink` need to return something other than `Unit`.

fdietze:
What are the blockers to merge this PR right now?

cornerman:
> What are the blockers to merge this PR right now?

@fdietze I think, the PR is good in general - I would love to iterate on the naming still as it was also discussed in the previous comments. Just never came around doing the renaming and am not sure what name would be right :/

busti:
I am not all that happy with outwatch having it's own streaming library implementation, as opposed to using *Tagless-Final* everywhere...  
I personally think it makes the internal code a bit harder to read overall and I am having trouble wrapping my head around the more complex parts. The mixing of streaming-lib type-classes and implementations in the same package is also a bit confusing, but that might also be fixable with better naming.  
But I get where it is coming from, it simplifies things a lot.

cornerman:
> I am not all that happy with outwatch having it's own streaming library implementation

@busti Me neither, one way would be to separate the typeclasses and the implementation into two different packages/libraries, so it becomes a bit more clear. But right now, especially the emitterbuilder depends on the implementation directly.

For the tagless final approach, I think it would be good, but I would want it to work very easily without having to provide the types explicitly all the time (if you only depend on one streaming library that might work with implicit resolution, but probably won't if you depend on multiple). I am totally open to a proposal how that could work :)

Otherwise, I will want to continue working on this PR, but need to reserve some time for it.

fdietze:
And as far as I understand, the streaming library so lightweight, that it is basically just a good management of callbacks, right?

cornerman:
> And as far as I understand, the streaming library so lightweight, that it is basically just a good management of callbacks, right?

Mainly yes, but of course it evolved a bit to give some nice sugar to operations you normally do in your app - like sampling, interval, etc. It is a bit similiar to what monix provides on the observable interface but less complex and only target for scala-js and without backpressure. In our benchmarks, it tends to be faster than monix.

reactive/src/main/scala/outwatch/reactive/Sink.scala
cornerman: You are right, the same holds for the `subscribe` method on the `Source`. It should return an effect type.

busti:
@cornerman This branch has come out of date with other outwatch-close dependencies, i.e. vectory, which causes it to fail to build in my project.  
I have reverted back to `master` for the time being, but so far I have not encountered any issues when using it.

cornerman:
@busti Good to know, thank you. Sorry for the inconvenience; I thought, the PR would be finished faster. But you never know :)

cornerman:
@busti I have rebased the branch on top of master again - so it should hopefully work again for the dependencies!

cornerman:
I have updated the PR to depend on the new library [`colibri`](https://github.com/cornerman/colibri`) that I have written. The readme and first comment in this PR are now updated.

PR #394, opened 2020-03-07
Boolean VDomModifier operations

cornerman:
Add `VDomModifier.ifTrue` and `VDomModifier.ifNot`

Based on #355


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: scala
 
 scala:
-  - 2.11.12
+  # - 2.11.12
```
fdietze: why are we dropping 2.11?

PR #399, opened 2020-03-11
Travis: use ScalaJS 0.6.32 (like in plugins)

fdietze:
Travis was running the old version...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ jdk:
   - openjdk11
 
 env:
-  - SCALAJS_VERSION=0.6.28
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,7 +9,7 @@ jdk:
   - openjdk11
 
 env:
-  - SCALAJS_VERSION=0.6.28
+  - SCALAJS_VERSION=0.6.31
```
cornerman: Did you mean `0.6.32` here?

fdietze: yes :D

PR #408, opened 2020-03-23
Library Helpers for Javascript Libraries

cornerman:
This PR adds helpers for javascript libraries:

- hammer-js
- clipboard-js
- *more to come*

I would like to add more that I already have implemented in my own projects. Just unsure whether this should live in this repository or in another repository. What do you all think?


busti:
I personally feel like this should live in a different repo. 
I'd go so far as to say that there should be a separate artifact for every supported js library, maybe even a separate repo each.

cornerman:
Yeah agree, we definitely need a separate artifact for each adapter library. I was a bit lazy but you are probably rightabout a different repo. I will make an outwatch-libs repo, where i will put all of them for now. We can think about whether there should be multiple repos :)

busti:
Sounds good. I'd be happy to contribute some of my own helpers.

cornerman:
Awesome! I have created a new repo (https://github.com/OutWatch/outwatch-libs) and will push these changes to it later.

PR #426, opened 2020-06-25
upgrade to scala.js 1.x

busti:
For with the sjs `0.6.x` support completely removed.  
There have been some breaking changes in `1.x`, but just minor ones affect us.  
I believe that we can drop support, since the lib is relatively feature complete and there isn't lots of new stuff on the table right now. Thoughts?  

Additionally, this PR depends on https://github.com/cornerman/colibri and has to be updated when that goes through.


busti:
The naming of either the parameters or the variable is poor. Which should be renamed?

outwatch/src/main/scala/outwatch/interpreter/SnabbdomOps.scala
```diff
@@ -117,7 +117,7 @@ private[outwatch] object SnabbdomOps {
 
       var proxy: VNodeProxy = null
       var nextModifiers: js.UndefOr[js.Array[StaticVDomModifier]] = js.undefined
-      var prependModifiers: js.UndefOr[js.Array[StaticVDomModifier]] = js.undefined
+      var _prependModifiers: js.UndefOr[js.Array[StaticVDomModifier]] = js.undefined
```
busti, line 120: The naming of either the variable or the parameters is poor. Which should be renamed?

cornerman: Why do we need to rename them?

cornerman:
This is great. Thank you so much! It will fix #396 

I have merged the PR in colibri!

Can you also update the badge in the readme? It should link to this svg: `http://www.scala-js.org/assets/badges/scalajs-1.0.0.svg`.

outwatch/src/main/scala/outwatch/interpreter/SnabbdomOps.scala
```diff
@@ -74,13 +74,25 @@ private[outwatch] object SnabbdomOps {
      }
    }
 
+   // See: https://www.scala-js.org/doc/interoperability/global-scope.html#dynamically-lookup-a-global-variable-given-its-name
```
cornerman, line 77: Maybe we can move this into the helpers package?

busti:
Ah, looks like the further updates on colibri caused some issues. Just a sec.

busti:
"a sec" turned into 4 hours, but all the tests succeed on my machine now. Let's just hope they also do on travis.  
Anyways, I updated all the tests to use the new scala-js-dom names and I replaced all occurrences of `initEvent` with the new event constructors.  

The only thing that's potentially missing is to create deprecations for the old attribute names.

busti:
I have created deprecations for the renamed attributes and tags.  
While we are at it, I think the ``` `class` ``` and ``` `for` ``` also need some rework, but that's something for another PR.

outwatch/src/main/scala/outwatch/interpreter/SnabbdomOps.scala
busti: I do not know why they cannot have the same name as the call-by-name params, but it would not let me compile if they had the same name.

cornerman: You mean the named arguments? I also do not understand. Will leave it like this for now then. I will try to find out what is happening :)

cornerman:
I have to admit the `idAttr` really does not look as good as the `id` before. But we'll see :)

PR #441, opened 2020-09-09
WIP: Modifier with Environment

cornerman:
TODO:
- [x] Environment for Modifier
- [x] Environment for EmitterBuilder
- [x] Render instances use Env (enables Kleisli rendering)
- [x] In companion objects of type aliases for all *Modifier[Any]: methods should have no type args
- [x] In companion objects of type aliases for all *EmitterBuilder[Any]: methods should have no type args
- [x] `provide` override for VNode instead of Outwatch.render* overloads

fixes #439, fixes #443, fixes #253

PR #442, opened 2020-09-09
WIP: ZIO Integration

cornerman:
Add another project `outwatch-zio` for better zio usage. It already works with zio thanks to cats-effect typeclasses, but you always need an implicit `Runtime` in scope, which is tedious to work with. The idea here is to construct an integration that uses a Modifier-Environment `Modifier[Has[Platform]]` (using ZLayer). Thereby we do not need any implicit in scope and only provide the platform when rendering the outwatch component. The platform is needed for the effectful modifiers to be run, when mounting the elements.

based on #441 

TODO:

- [x] Render instance without implicit runtime
- [ ] EmitterBuilder integration without implicit runtime
- [ ] Observable integration without implicit runtime. Colibri?
- [ ] ZStream instances for Source/Sink. Colibri?


busti:
I like this approach. As long as the core of the lib stays generic, I am happy.

cornerman:
> I like this approach. As long as the core of the lib stays generic, I am happy.

Yes! The core stays totally generic and only depends on cats/cats-effect :)

cornerman:
Now tracked in #441

PR #481, opened 2021-01-04
Removed obsolete module reference: snabbdom/modules/class

abdolence:
Thanks for your work.

The snabbdom class module was removed recently (https://github.com/OutWatch/snabbdom/commit/b6cb8ecffab4c9ef757be172529b194b87a9707d), 
but it still exists in Scala facade.


cornerman:
Thank you very much! That should indeed be removed.

I just need to update outwatch to the newest snabbdom-outwatch release. Then I will be merge this one :)

PR #540, opened 2021-11-06
Update to scalajs-dom 2.0.0 and domtypes 0.15.0

armanbilge:
Supersedes https://github.com/outwatch/outwatch/pull/534 and https://github.com/outwatch/outwatch/pull/539. Also bumps colibri to the latest hash.

This is almost ready, except for some trickyness I will point out and would appreciate some feedback on. Thanks!

### DIFF
--- a/outwatch/src/main/scala/outwatch/definitions/OutwatchAttributes.scala
+++ b/outwatch/src/main/scala/outwatch/definitions/OutwatchAttributes.scala
@@ -125,7 +125,7 @@ trait AttributeHelpers { self: Attributes =>
 
   @inline def attr[T](key: String, convert: T => Attr.Value = (t: T) => t.toString : Attr.Value) = new BasicAttrBuilder[T](key, convert)
   @inline def prop[T](key: String, convert: T => Prop.Value = (t: T) => t) = new PropBuilder[T](key, convert)
-  @inline def style[T](key: String) = new BasicStyleBuilder[T](key)
+  @inline def style[T](key: String, @annotation.nowarn dummy: Unit = ()) = new BasicStyleBuilder[T](key)
 
   @inline def emitter[F[_] : Source, E](source: F[E]): EmitterBuilder[E, VDomModifier] = EmitterBuilder.fromSource(source)
 
--- a/util/src/main/scala/outwatch/util/Http.scala
+++ b/util/src/main/scala/outwatch/util/Http.scala
@@ -70,7 +70,7 @@ object Http {
     headers = request.headers,
     withCredentials = request.withCredentials,
     responseType = request.responseType
-  )
+  ): @scala.annotation.nowarn("cat=deprecation") // TODO
 
   private def request(observable: Observable[Request], requestType: HttpRequestType)(implicit ec: ExecutionContext): Observable[Response] =
     observable.switchMap { request =>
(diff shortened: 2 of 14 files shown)


util/src/main/scala/outwatch/util/Http.scala
```diff
@@ -70,7 +70,7 @@ object Http {
     headers = request.headers,
     withCredentials = request.withCredentials,
     responseType = request.responseType
-  )
+  ): @scala.annotation.nowarn("cat=deprecation") // TODO
```
armanbilge, line 73: The AJAX helpers in scala-js-dom are deprecated. It'll take a bit of work to rework it here, ok to leave to a future PR?

armanbilge:
CI/compilation currently fails because this method:
https://github.com/outwatch/outwatch/blob/3ac5a07311e84a3727bc5e86fa3fdba5a833ff16/outwatch/src/main/scala/outwatch/definitions/OutwatchAttributes.scala#L128

conflicts with this method from scala-dom-types:
```scala
  @inline protected def style[V](key: String): Style[V] = new Style(name = key)
```
https://github.com/raquo/scala-dom-types/blob/f8a35f28703cd7fb33ad444b6873238967caf98a/shared/src/main/scala/com/raquo/domtypes/generic/builders/StyleBuilders.scala#L15

And they are both inherited in:
https://github.com/outwatch/outwatch/blob/3ac5a07311e84a3727bc5e86fa3fdba5a833ff16/outwatch/src/main/scala/outwatch/dsl.scala#L5

Any ideas how to resolve this? Note that we have an implicit conversion to convert between the two types:
https://github.com/outwatch/outwatch/blob/3ac5a07311e84a3727bc5e86fa3fdba5a833ff16/outwatch/src/main/scala/outwatch/package.scala#L8

Thanks in advance!

util/src/main/scala/outwatch/util/Http.scala
cornerman: I think, that is fine for now :+1:

cornerman:
@armanbilge I have added a PR to your fork against this branch, that fixes these errors. I think, we can merge it then :)

armanbilge:
Thanks for the assist!

cornerman:
Awesome. Thank you!

cornerman (approved):
LGTM :+1:

outwatch/src/main/scala/outwatch/definitions/OutwatchAttributes.scala
```diff
@@ -125,7 +125,7 @@ trait AttributeHelpers { self: Attributes =>
 
   @inline def attr[T](key: String, convert: T => Attr.Value = (t: T) => t.toString : Attr.Value) = new BasicAttrBuilder[T](key, convert)
   @inline def prop[T](key: String, convert: T => Prop.Value = (t: T) => t) = new PropBuilder[T](key, convert)
-  @inline def style[T](key: String) = new BasicStyleBuilder[T](key)
+  @inline def style[T](key: String, @annotation.nowarn dummy: Unit = ()) = new BasicStyleBuilder[T](key)
```
armanbilge, line 128: Clever! I will add it to my bag of tricks :)

armanbilge:
Hooray!

PR #542, opened 2021-11-19
gh-actions build also checks out git tags

FloWi:


.github/workflows/build.yml
```diff
@@ -14,6 +14,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
cornerman, line 18: I think, we only need it in the publish job, right?

FloWi: I also added it here to avoid confusion when staring at the logs. But I saw that the build job doesn't log the version.

cornerman (approved):
:+1:

PR #563, opened 2022-01-05
Update to Scala.js 1.8.0

armanbilge:
Supersedes https://github.com/outwatch/outwatch/pull/551. This should hopefully get you unblocked while we wait for https://github.com/scalameta/mdoc/pull/597.

PR #564, opened 2022-01-08
Fix SJS 1.8 update

armanbilge:
Follow up to https://github.com/outwatch/outwatch/pull/563#issuecomment-1007982347.

This is ... not nice. I don't think it would be unreasonable to rollback to SJS 1.7 instead. Whatever works for you 👍

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -205,10 +205,16 @@ lazy val docs = project
       "REPOURL" -> "https://github.com/OutWatch/outwatch/blob/master",
       "js-mount-node" -> "docPreview"
     ),
-    libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


cornerman:
Thank you! It is not nice, but at least fixes the build :)

Either we wait for an update of mdoc or will revert afterwards.

armanbilge:
FWIW I'm not hopeful about an mdoc fix coming soon. Problem (1) in https://github.com/scalameta/mdoc/issues/601 may not be so easy to fix.

Things will work again for a while after Scala 3.2 is released and a new mdoc based on that, until the next SJS update breaks it again.

PR #577, opened 2022-02-11
accum for style attributes

### LINKED ISSUE #212: Style should be Accumulating by default
We can currently write `div(style := "background:white", style.accum := "color:red)"` if we want multiple style attributes to accumulate. For the `class` attribute, accum is already default. I think, it should be default for `style` as well.

cornerman:
fixes #212


fdietze:
A test case would be good.

cornerman:
Good point. Done!

fdietze (approved):

PR #582, opened 2022-02-23
Update docs with correct port for template

JPonte:
It's just fixing the port. The large amount of commit lines is just GitHub normalizing the line endings of the file it seems.


cornerman:
Good catch! Thank you for this PR.

There is just a bit more changes in the diff than just the port. I had the same problem once when editing on github - where it was adding line-end changes. Can you try to reduce the diff? :)

cornerman (approved):
Thank you :)

PR #591, opened 2022-03-28
refactor localstorage interface

cornerman:
Please just check the last commit.

Based on #586


cornerman:
Moved to colibri. The deprecation is now part of #586 

Use `colibri.jsdom.Storage.Local` or `colibri.jsdom.Storage.Session`.

PR #605, opened 2022-04-04
Async patching and batching

cornerman:
The main change is to move snabbdom patching into a micro task (currently running synchronously). Now it would run async but before any rendering or dom events. See about event loops in javascript: https://javascript.info/event-loop

I owe this idea to @armanbilge. Thank you so much :)

The advantages:
- We will gather all observable triggers (even from nested observables) before the first render -> this circumvents triggering multiple succeeding patch calls inside one transcation.
- Multiple observables inside one node only do one patch call!
- In diamond cases, we will not render invalid state -> no (visible) frp glitches!
- Promise-based modifiers can still be captured inside the initial render and do not need an additional patch.
- When triggering from an emitter builder, we do not render inside the event callback - which becomes potentially long-running.

The disadvantages:
- Had to rewrite a lot of tests 😵‍💫

### DIFF
--- a/outwatch/src/main/scala/outwatch/helpers/MutableNestedArray.scala
+++ b/outwatch/src/main/scala/outwatch/helpers/MutableNestedArray.scala
@@ -13,6 +13,12 @@ private[outwatch] final class MutableNestedArray[T] {
     case t: T@unchecked => f(t)
   })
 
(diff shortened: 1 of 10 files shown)


cornerman:
@mergifyio rebase

cornerman:
@Mergifyio rebase

PR #617, opened 2022-05-30
Add Scalafmt

### LINKED ISSUE #452: Style guide/rules
~~We have a scalastyle config and reference it in `Contributing.md`, but do not actually use it.~~ We should/could actually use a formatter. Or we decide to not use a formatter and remove it from the contributing guide.

fdietze:
closes #452


busti:
Does rhis PR already apply scalafmt to the entire codebase?  
I cannot believe that there would be no formatting changes

busti:
Yeah, that makes sense. The refactoring would be painful.

PR #657, opened 2022-10-21
deprecate AttributeBuilder#toggle and introduce new variant when

### LINKED ISSUE #641: cls.toggle is not very nice to use.
I feel like `cls.toggle` is in stark contrast to other class definitions in outwatch.  
Take a look at the following code:  
```scala
val disabled: Rx[Boolean] = ???

button(
  cls := "btn",
  cls.toggle("disabled") <-- disabled
)
```

It feels like this should be the other way round. `cls.toggle(disabled) := "disabled"`.

cornerman:
fixes #641


cornerman:
The toggle method is deprecated with PR #659 without replacement.

I think, we should leverage Observable methods. I have added `toIf` in colibri for this use-case:
```boolObservable.toIf[VModifier](cls := "disabled")``` or ```boolObservable.toIf(cls := "disabled", display.none)```

PR #672, opened 2022-12-05
make EmitterBuilder extend AttrBuilder

cornerman:
This allows assigning a function to an emitter like:
onClick := { event =>
  console.log(event)
}


fdietze:
This change should come with a test for future compatibility.

PR #675, opened 2023-01-04
feat: add implicit to render show instances

busti:
Also includes a test for the implicit order.  

Unfortunately tests won't run on my machine since `node-gyp` is acting up as per usual, so I may need to make some changes to the implicit order once the github tests are done.


cornerman (approved):
Nice, good idea!

PR #694, opened 2023-02-22
rename EmitterBuilder#discard to EmitterBuilder#done

cornerman:
This method name comes from akka-streams. I find it more intuitive than `discard` - to me `discard` sounds like discarding the whole stream. Though, you can also interpret it as discarding the value in the stream. On the other hand, `done` clearly means that you are finished with the processing of the stream.


fdietze (approved):
To me, `discard` was intuitive in meaning it discards the value, but still runs the stream. But I'm fine either way. If it is more consistent with other libraries in the scala ecosystem, we should do it.

PR #696, opened 2023-02-22
Add an optional config to RenderConfig to allow for 3rd party libraries to hook into attr merging

busti:
This Pull Request adds a hook into `RenderConfig` which allows third party libraries to hook into attribute merging. 

This is especially useful for class names which can override each other, like those used by tailwind. 
This feature could for example be used with `tailwind-merge` as follows:  

```scala
import outwatch._
import outwatch.dsl._
import tailwindMerge._

val config = RenderConfig(
  errorModifier: _ => VMod.empty,  
  accumAttrHook => Some((attr, other) => if (attr.title == "class") twMerge(attr.value, other) else a.accum(attr.value, other)),
)

Outwatch.renderInto("#app", div(cls := "hidden", cls := "visible", config)
```


cornerman:
@busti Sorry, just realized my review comment was pending the whole time and was not sent out :/

cornerman (approved):
Thank you. LGTM :+1:

busti:
Had to rename the parameter since scalac was complaining that it was shadowing the `a` VNode from the dsl.

busti:
For some reason the tests still fail though.

busti:
Now just about everything fails. 
Also I apparently misinterpreted the error message from before.  
Problem is I cannot reproduce the original error locally.

PR #727, opened 2023-08-16
remove Render instances for cats.Show and add Render.show[T]

cornerman:
We cannot assume that a Show instance is always meant to be rendered in the UI.


cornerman:
An example: `Show[Unit].show(()) == "()"`.

fdietze (approved):

PR #737, opened 2023-09-22
add ConfiguredModifier to change RenderConfig inside the virtual dom

cornerman:
Use-case: Catch exception inside a subtree of the virtual dom.

You can now change the `RenderConfig` for a modifier, and thereby also change the `RenderConfig#errorModifier`.

Usage:

```VMod.configured(myModifier)(outsideRenderConfig => outsideRenderConfig.copy(...))```

```VMod.catchError(myModifier)(throwable => div("broken " + throwable))```


futile:
Awesome, thanks a lot, I _think_ this is exactly the issue we stumbled upon :) Also tagging @busti :)

busti:
🥳 That was surprisingly less complicated than I imagined it would be

